PR #106, opened 2022-04-27
fix: set search filter searchFor initialValue

LorenzoArkemis:
### Description

- set searchFor initial value to fiscalCode in FilterNotificationsTable.tsx
- change test accordingly


LorenzoCocciarelli:
Wrong user

PR #158, opened 2022-05-26
Feature/pn 1364 fe rivedere il layout generale di pn logica margini padding

maurizioflauti:
refactor: change layout structure

add sticky footer using min-height and flex-grow instead of calc()

change sidebar and main container structure and style
apply layout and style changes to every page in order to make them
appear consistent one each other


packages/pn-personafisica-webapp/src/pages/NuovaDelega.page.tsx
```diff
@@ -4,7 +4,6 @@ import { useTranslation } from 'react-i18next';
 import {
```
AndreaCimini90, line 4: Verificare con Andrei o Lorenzo

AndreaCimini90 (changes requested):

PR #278, opened 2022-07-05
missed localizations

AndreaCimini90:


ZorbaDimatteo (changes requested):
@AndreaCimini90 in locale mi fallisce la loginError

PR #415, opened 2022-10-03
Fix/pn 2210 add usermap function in auth

LorenzoCocciarelli:
Token exchange response gets now mapped to user fields


LorenzoCocciarelli:
Restored spread of exchange token response in user fields in packages/pn-pa-webapp/src/api/auth/Auth.api.ts and packages/pn-personafisica-webapp/src/api/auth/Auth.api.ts

PR #436, opened 2022-10-18
PN-2261 Handling failure in API calls to retrieve information, for pa-webapp

clombardi:
## Short description
Use in pa-webapp the tools to handle the cases in which an API call made to fetch information fails, i.e. yields a status code in the 4xx or 5xx ranges.
In most cases, some part of the page is replaced by a banner which indicates the inability of fetching data, and allows the user to retry the data fetch. In the particular case of the call which fetches the ToS accept status, a failure implies an immediate logout.

- Apply the `<ApiErrorWrapper>` or `<ApiError>` components, in the latter case with ad-hoc handling of the API call error (obtained through the `useErrors` custom hook, for the API calls regarding: fetch of a list of notifications, fetch of the details of one modification, fetch of the user groups used to create a notification, fetch of the organization party corresponding to the logged user. In the latter case, the `ApiErrorWrapper` is used inside the router, an ad-hoc guard was implemented to this end.
- Use the `performThunkAction` to simplify the development of some Redux thunk actions, and to avoid forgetting to soundly manage the eventual exceptions.
- Added a `renderWithoutRouter` in `src/__test__/test-utils.tsx`, it simplifies the writing of tests for which we want to set an initial Redux store state, but we want to use a react-router Router ad-hoc for the test, and not the `BrowserRouter` defined in the `render` function inside the same file.

## How to test
You should be able to simulate a failure in one API call. The involved calls are:
- `getOrganizationParty` (in ExternalRegistries), this is invoked when entering the app.
- `getSentNotifications`, to obtain the list of notifications.
- `getSentNotification`, to obtain the detail of one notification.
- `getUserGroups`, to obtain the options for the group select in the workflow of manual creation of a notification, in the first step.
- `getConsentByType` (in Consents), to obtain the ToS acceptance status.

The next-to-last commit includes mock versions for several of these API calls.


packages/pn-pa-webapp/src/pages/components/NewNotification/PreliminaryInformations.tsx
```diff
@@ -83,128 +83,139 @@ const PreliminaryInformations = ({ notification, onConfirm }: Props) => {
     trackEventByType(TrackEventType.NOTIFICATION_SEND_PAYMENT_MODE, { target: e.target.value });
   };
 
-  useEffect(() => {
+  const fetchGroups = useCallback(() => {
     if (groups.length === 0) {
       void dispatch(getUserGroups(GroupStatus.ACTIVE));
     }
-  }, []);
+  }, [groups]);
```
AndreaCimini90, line 90: There are cases in which groups can be empty. This condition creates an infinite-loop.
The groups.length condition can be removed and the useCallback dependencies too.

packages/pn-pa-webapp/src/navigation/routes.tsx
```diff
@@ -8,31 +10,60 @@ import NewNotification from '../pages/NewNotification.page';
...
-           * Refers to PN-1741
-           * Commented out because beyond MVP scope
-           *
-           * LINKED TO:
-           * - "const BasicMenuItems" in packages/pn-pa-webapp/src/utils/role.utility.ts
-           * - BasicMenuItems in packages/pn-pa-webapp/src/utils/__TEST__/role.utilitytest.ts
-           *
-           * <Route path={routes.API_KEYS} element={<ApiKeys />} />
-           * */}
+        <Route path="/" element={<OrganizationPartyGuard />}>
```
AndreaCimini90, line 49: We can move the logic of OrganizationPartyGuard into App.tsx?

AndreaCimini90 (changes requested):

packages/pn-pa-webapp/src/pages/components/NewNotification/PreliminaryInformations.tsx
clombardi: OK, I remove both the condition and the dependency.
I usually include in the dependency list all the data items which are read inside the function, since React documentation indicates to do so. In fact, in other projects where I worked before, a warning arises whenever some read datum is not included in a dependency list.
In this particular case, once the condition is removed, the dependency can be removed as well since groups is no longer read. But I still mention that in other points in the pn FE codebase, there are data items which are read in a useCallback / useEffect / useMemo / etc. function, and which are not included in the dependency list.

packages/pn-pa-webapp/src/navigation/routes.tsx
clombardi: I would prefer to keep OrganizationPartyGuard as a proper guard, since it does not affect all the routes, there are indeed public routes (currently PRIVACY_POLICY and TERMS_OF_SERVICE, but I imagine others could be added later) where this guard shouldn't apply. So I keep the code in its current form, at least until I receive a subsequent review.

AndreaCimini90: OK per mantenere il guard ma lo sposterei in file e scriverei il test relativo

clombardi: I did as requested, albeit I don't agree with the comment. The decision of keeping `OrganizationPartyGuard` inside `routes.tsx` was deliberate, as in my view, the component is (at least in principle) of no interest outside the `Router` definition, and the inclusion in the same file does not make the file too big or complex. In fact, `OrganizationPartyGuard` was not exported, and this decision conveys a meaning: the component is something particular to the stuff implemented in `routes.tsx`.  
Besides this, I don't agree to separate each React component in a separate file as a general criterion, the ability to include several definitions in the same class is an aspect of Javascript that actually attracts me in comparison with (the boring) Java.  
I remark that the case which raised the convenience of defining `OrganizationPartyGuard`, namely when the API call to getOrganizationParty yields an error, was already covered by a test in `routes.test.ts`.

In general, I guess that it could be convenient to think about the thin line between a design flaw and the personal preferences of each team member.

AndreaCimini90 (approved):

PR #442, opened 2022-10-24
Feat/pn 2298 add tos guard

LorenzoCocciarelli:
## Short description
Introduces ToSGuard and remove tos as route, both in pa and personafisica.

- Adds ToSGuard in routes
- Adds tests

## How to test
If the user has not accepted the tos, the tos acceptance page should be displayed, otherwise the current page should be displayed.
The route should not change
While fetching the tos value the loading page should be displayed.


packages/pn-pa-webapp/src/App.tsx
```diff
@@ -159,6 +160,10 @@ const App = () => {
       ),
   });
 
+  useEffect(() =>{
```
AndreaCimini90, line 163: Facendo delle prove, al primo login si ha un errore in quanto viene chiamata la tos senza essere autenticato.
Non si può spostare la getTosApproval nel TosGuard? Se si applicare lo stesso cambiamento a persona fisica

AndreaCimini90 (changes requested):

LorenzoCocciarelli:
Description:
- Aligned position of ToSGuard between pn-pa-webapp e pn-personafisica-webapp
- Fixed merge conflict
- Fixed tests

AndreaCimini90 (approved):

PR #443, opened 2022-10-25
boolean and optional keys management

AndreaCimini90:
## Short description
Now pn-validator works with boolean and optional properties

- Modified TypeRules to manage this two cases

## How to test
Create a class with boolean and optional properties. Create validator for this class and check that ide suggests correct rules for these properties

### DIFF
--- a/packages/pn-validator/src/types/TypeRules.ts
+++ b/packages/pn-validator/src/types/TypeRules.ts
@@ -4,20 +4,41 @@ import { DateRuleValidator } from './../ruleValidators/DateRuleValidator';
 import { ObjectRuleValidator } from '../ruleValidators/ObjectRuleValidator';
 import { ArrayRuleValidator } from './../ruleValidators/ArrayRuleValidator';
 import { CommonRuleValidator } from '../ruleValidators/CommonRuleValidator';
+import { Validator } from '../Validator';
 
-type MixedType = String | Number | Date | Boolean | undefined | null;
+type MixedType = String | Number | Date | Boolean;
 
-export type TypeRules<TModel, TValue> = (TValue extends String | undefined | null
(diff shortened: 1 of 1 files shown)


clombardi (approved):
OK, I tested with the following code

```
import { Validator } from "@pagopa-pn/pn-validator";

export interface BEIncident {
  functionality: string;
  status: string;
  startDate: string;
  endDate?: string;
  legalFactId?: string;
  fileAvailable?: boolean;
}

function validateIsoDate(required: boolean) {
  return (value: string | undefined) => {
    const isOK = value
      ? /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2}(?:\.\d{1,3})?)Z$/.test(value) && !Number.isNaN(Date.parse(value))
      : !required;
    return isOK ? null : "A date in ISO format is expected";
  }
}

function validateString(value: string | undefined): string | null {
  const isOK = value === undefined || typeof value === 'string';
  return isOK ? null : "A string is expected";
}

function validateBoolean(value: boolean | undefined): string | null {
  const isOK = value === undefined || typeof value === 'boolean';
  return isOK ? null : "A boolean is expected";
}


class BEIncidentValidator extends Validator<BEIncident> {
  constructor() {
    super();
    this.ruleFor('functionality').customValidator(validateString).isEmpty(true);
    // this.ruleFor('status').isOneOf(Object.values(IncidentStatus) as string[]);
    this.ruleFor('startDate').customValidator(validateIsoDate(true));
    this.ruleFor('endDate').isEqual('toto');
    this.ruleFor('legalFactId').customValidator(validateString);
    this.ruleFor('fileAvailable').customValidator(validateBoolean);
  }
}
```

(the commented line uses a rule I will integrate in a later PR), it compiled perfectly, so in my opinion it would be desirable to make this PR be merged into develop in order to use this version of pn-validator.

PR #453, opened 2022-11-04
PN 1953 - FE show status and downtime log, allow to download downtime certification

clombardi:
## Short description
This PR includes all the needed stuff to show a page of app status and downtime log, including the possibility of downloading a legally valid certificate ("attestazione opponibile a terzi") for any downtime event.

- We define a model for the app status and for each downtime event.
- We implement the API calls for the status/downtime API provided by DXC, including validations performed through `pn-validator`. Is the first actual use of `pn-validator` in operative code. There are several tests regarding different possible cases of corrupted data coming from API. 
- We implemented the AppStatus page. We note the way in which we avoided duplication of logic between the mobile and desktop versions of the downtime log table.
- We enabled the use of the `pn-validator` package in both pf-webapp and pa-webapp, by adding the corresponding line in the `craco.config.js` files in each folder.
- We made some modifications to pn-common components
  - `ItemsCard` (used to show a list of elements in mobile mode), now the header can include either one or two elements. For the list of downtimes, we needed just one element in header, centered. To center the header we use the already available `headerGridProps` prop.
  - Also in `ItemsCard`, we fixed some bugs, specifically the calls to `getLabel` didn't pass the whole `Item` as parameter, so any definition of `getLabel` which actually needed the `Item` raised an error.
  - Still in `ItemsCard`, we added `data-testid="cardBodyValue"` for values not wrapped in typography, to make all values alike to this respect (previously, only the values wrapped in typography by `ItemsCard` carried this `data-testid`).
  - `EmptyState`: refactored the specification of which "sentiment" (sad / happy / custom) icon should be shown. The previous `disableSentimentDissatisfied` prop was replaced with a new `sentimentIcon`, which can be either a member of the `KnownSentiment` enum (options DISSATISFIED / SATISFIED / NONE) or a `SvgIconComponent`. In this way, we promote uniformity for the common cases, and at the same time, allow for arbitrary "sentiment icons". The default is `KnownSentiment.DISSATISFIED`, to preserve the default behavior (`disableSentimentDissatisfied = false`) of the previous version. <br/> I changed and manually tested all the uses of `EmptyState` in pa-webapp and pf-webapp.
  - Also in `EmptyState`, we adjusted the size of the "face" icon, either happy, sad or custom; this was agreed with the UI team (Norma and Alice).
  - `SideMenuListItem`: now the icon can be either a React component name (as up to now), or a function which yields the already instantiated icon. To achieve this, I changed the `SideMenuItem` interface. We needed this because the icon to show in the side menu for the app status option (as well as its color) depends on the actual status (OK or not-OK).
  - date services/utilities: we added two functions to format date/time, and an additional utility function. 
    - The new function `formatDateTime` yields an object `{ date, time }` with the strings that represent the corresponding parts of a timestamp; the time has the legend _"ore "_ prepended, e.g. _"ore 15:08"_. This legend _"ore"_ is i18n-ed.
    - In turn, `formatTimeHHMM`, yields the hour/minute part of a Date in HH:MM format, 
    - Finally, we also added a `minutesBeforeNow` function which yields a date in the past ... as the function name indicates.
  - Also regarding date services, the _"Oggi"_ legend issued by `formatDate` was i18n-ed.
  - Data regex: we added the `isoDate` regex.
  - We added a new `useDownloadDocument` hook that takes in charge the download of a document, given its URL. We used this hook in the new app status page. It could be used as well for the various downloads in the notification detail FE, I opened https://pagopa.atlassian.net/browse/PN-2708 to this respect.
- As indicated above, we actually implemented the first use of the `pn-validator` package. Some modifications of that package were needed. We also add the `IsOneOf` rule as part of this PR.

## How to test
Just launch the app (at the moment only pf, I should replicate the FE in pa next week), and look for a "Stato della piattaforma" option in the left-hand side menu. In DEV environment, this uses an actual API provided by DXC.  
If you want to variate the results, you can play with `AppStatus.api.ts`, there is a `useMockResponseData` boolean, just enable it for development (for test should be always false), the mock responses are in the same `AppStatus.api.ts` file.


clombardi:
I finally finished the work on this FE. It is possible that the checks report a failure in the tests of `pn-personafisica-login`. Unfortunately, I cannot reproduce nor understand the problem, since the tests pass when I run them locally on my PC, and I cannot access to the AWS logs.

packages/pn-commons/src/components/AppStatus/downtimeLog.utils.tsx
```diff
@@ -0,0 +1,136 @@
...
+        <Typography variant="body2">{dateAndTime.date},</Typography>
+        <Typography variant="body2">{dateAndTime.time}</Typography>
+      </Stack>
+      : <Typography variant="body2">{dateAndTime.date}, {dateAndTime.time}</Typography>;
+  } else {
+    return <Typography variant="body2">-</Typography>;
+  }
+};
+
+export function adaptFieldSpecToMobile(desktopFieldSpec: Omit<Column<DowntimeLogColumn>, "width">): CardElement {
```
AndreaCimini90, line 28: Maybe in the future we can standardize the label field between Desktop and Mobile and remove this function.
Now we can create an issue for this

packages/pn-commons/src/components/AppStatus/downtimeLog.utils.tsx
```diff
@@ -0,0 +1,136 @@
...
+  | 'functionality'
+  | 'legalFactId'
+  | 'status'
+  | '';
+
+
+export function useFieldSpecs({ getDowntimeLegalFactDocumentDetails }: { getDowntimeLegalFactDocumentDetails: (legalFactId: string) => any}) {
+  const theme = useTheme();
+
+  const getDateFieldSpec = useCallback((fieldId: DowntimeLogColumn, inTwoLines: boolean): Omit<Column<DowntimeLogColumn>, "width"> => ({
```
AndreaCimini90, line 49: Because we return object, I would replace useCallback (for functions) with useMemo (for variables).
This for all useCallback below

packages/pn-commons/src/models/appStatus.ts
```diff
@@ -0,0 +1,173 @@
...
+ * - functionality not in the expected set 
+ *   (if possible, taken from the response from /downtime/v1/status)
+ *   but in order to verify it, I should be able to access the Redux store
+ *   (to avoid re-fetching the set of functionalities each time downtimes are retrieved).
+ *   Hence I won't validate, and rather indicate "unknown functionality" in the FE.
+ * - status not in the DowntimeStatus set
+ * - startDate not a valid date
+ * - endDate, if present, not a valid date
+ */
+ export interface BEDowntime {
```
AndreaCimini90, line 93: I would replace BExxx with xxxDTO.
The same for the interfaces below

packages/pn-commons/src/components/EmptyState.tsx
```diff
@@ -11,6 +11,8 @@ type Props = {
   emptyActionLabel?: string;
   /** Disable sad emoticon */
   disableSentimentDissatisfied?: boolean;
+  /** Enable happy emoticon */
+  enableSentimentSatisfied?: boolean;
```
AndreaCimini90, line 15: Maybe it would better if it is more generic. Instead of using a boolean, i would pass the icon directly

packages/pn-commons/src/services/date.service.ts
```diff
@@ -11,7 +13,28 @@ export function formatDate(dateString: string): string {
...
-  return isToday(date) ? 'Oggi' : `${day}/${month}/${date.getFullYear()}`;
+  const todayLabel = getLocalizedOrDefaultLabel(
+    'common',
+    'date-time.today-uppercase-initial',
+    "Oggi"
+  );
+  return isToday(date) ? todayLabel : `${day}/${month}/${date.getFullYear()}`;
+}
+
+export function formatTimeHHMM(dateString: string): string {
```
AndreaCimini90, line 24: Add tests for new functions

packages/pn-pa-webapp/src/App.tsx
```diff
@@ -38,26 +45,61 @@ const App = () => {
...
+       * - "<Route path={routes.API_KEYS}.../>" in packages/pn-pa-webapp/src/navigation/routes.tsx
+       * - BasicMenuItems in packages/pn-pa-webapp/src/utils/__TEST__/role.utilitytest.ts
+       */
+      // { label: menu.api-key, icon: VpnKey, route: routes.API_KEYS },
+      { 
+        label: t('menu.app-status'), 
+        // ATTENTION - a similar logic to choose the icon and its color is implemented in AppStatusBar (in pn-commons)
+        icon: () => currentStatus 
+          ? (currentStatus.appIsFullyOperative
+            ? <CheckCircleIcon sx={{ color: theme.palette.success.main }} />
```
AndreaCimini90, line 77: Check if we can avoid the use of useTheme(). When we use the material-ui components we can simply do this:
<Component  sx={{color: 'name of the color in the theme'}}/>

packages/pn-pa-webapp/src/pages/AppStatus.page.tsx
```diff
@@ -0,0 +1,145 @@
...
+        title={t('appStatus.title')} variantTitle='h4'
+        subTitle={t('appStatus.subtitle')} variantSubTitle='body1'
+      />
+
+      {/* Dati relativi al status */}
+      <ApiErrorWrapper apiId={APP_STATUS_ACTIONS.GET_CURRENT_STATUS} reloadAction={() => fetchCurrentStatus()} mt={3}>
+        {currentStatus && <AppStatusBar status={currentStatus} />}
+        { currentStatus &&
+          <Stack direction="row" justifyContent="center">
+            <Typography variant="caption" sx={{ mt: 2, color: theme.palette.text.secondary }}>
```
AndreaCimini90, line 101: Also here maybe we can avoid the use of useTheme

packages/pn-pa-webapp/src/pages/components/AppStatus/useDownloadDocument.ts
```diff
@@ -0,0 +1,28 @@
...
+  /* eslint-disable functional/immutable-data */
+  const link = document.createElement('a');
+  link.href = url;
+  link.target = '_blank';
+  link.rel = 'noreferrer';
+  link.click();
+  /* eslint-enable functional/immutable-data */
+};
+
+export function useDownloadDocument() {
```
AndreaCimini90, line 16: Maybe this can become a shared hook

packages/pn-personafisica-webapp/src/App.tsx
```diff
@@ -46,11 +50,13 @@ const productsList: Array<ProductSwitchItem> = [
 const App = () => {
   const dispatch = useAppDispatch();
   const { t, i18n } = useTranslation(['common', 'notifiche']);
+  const theme = useTheme();
```
AndreaCimini90, line 53: Same before

packages/pn-pa-webapp/src/utils/role.utility.ts
```diff
@@ -1,22 +1,10 @@
-import { Email, People, SupervisedUserCircle } from '@mui/icons-material';
+import { People, SupervisedUserCircle } from '@mui/icons-material';
 import { SideMenuItem } from '@pagopa-pn/pn-commons';
 
 import * as routes from '../navigation/routes.const';
 import { PNRole } from '../models/user';
 import { IS_DEVELOP } from './constants';
 
-const BasicMenuItems: Array<SideMenuItem> = [
```
AndreaCimini90, line 8: I'm not sure about moving this logic to App.tsx. We can manage here the sidemenu items and lighten App.tsx code

AndreaCimini90 (changes requested):

packages/pn-commons/src/components/AppStatus/downtimeLog.utils.tsx
clombardi: Created the issue https://pagopa.atlassian.net/browse/PN-2691

packages/pn-commons/src/components/AppStatus/downtimeLog.utils.tsx
clombardi: The four `get...FieldSpec`, as well as `getRows`, are indeed functions. Please check how they are used in `DesktopDowntimeLog.tsx` and `MobileDowntimeLog.tsx`. Therefore, I still guess that `useCallback` is the right hook for these definitions.

packages/pn-commons/src/models/appStatus.ts
clombardi: Done, the following replacement have been performed
- `BEDowntime` as `DowntimeDTO`
- `BEDowntimeLogPage` as `DowntimeLogPageDTO`
- `BEStatus` as `AppStatusDTO`
- `BEDowntimeLogPageValidator` as `DowntimeLogPageDTOValidator`
- `BEStatusValidator` as `AppStatusDTOValidator`

packages/pn-commons/src/components/EmptyState.tsx
clombardi: In my opinion, this depends on who should decide about the icon to show, either `EmptyState` or the component which calls it. 
If I (a component) want to use `EmptyState` with a happy face, what should I indicate, "show a happy face" or else "show this icon"? The former enhances uniformity (as the decision is on `EmptyState`, all happy faces will look the same), the latter enhances flexibility.
In fact, a similar argument could be made about the sad face.
I propose to replace both `disableSentimentDissatisfied` and `enableSentimentSatisfied` by a single `sentimentIcon` atrribute, whose type would be 'KnownSentiments | IconType`, where `KnownSentiments` would be an enum with two elements, DISSATISFIED and SATISFIED'. The default would be `KnownSentiments.DISSATISFIED`. What do you think?

packages/pn-commons/src/services/date.service.ts
clombardi: Added, please note that I had to localize the hour in order to build the expected result; cfr. the comments in the test.

packages/pn-pa-webapp/src/App.tsx
clombardi: Done, even when I personally prefer to use the theme, to avoid losing type discipline (`theme.palette.success.main` is checked by the editor and the compiler, while `"success.main"` is not).

packages/pn-personafisica-webapp/src/App.tsx
clombardi: Done, same comment applies

packages/pn-pa-webapp/src/pages/AppStatus.page.tsx
clombardi: done, both in pa-webapp and pf-webapp

packages/pn-pa-webapp/src/pages/components/AppStatus/useDownloadDocument.ts
AndreaCimini90: Maybe we can pass the legalFactDocumentDetails.url and the clearLegalFactDocumentData as props and macke the hook generic.
About the way we manage the downloads, do you have some advice?

packages/pn-pa-webapp/src/utils/role.utility.ts
clombardi: To define the basic menu items in a file about roles doesn't make sense to me. I would rather define a `useMenuItems` hook which (export a function that) resolves the items, including translation. Defining the function inside a hook eases the access to the Redux store and the translation facilities, so that no (somewhat arbitrary, at least in my opinion) parameters would be needed.  
Do you want me to implement such a hook? Btw, a similar solution could be applied in pf-webapp; there is a comment about move the menuItems definition to a utility function. Do you want me to apply a hook-based solution in both pa and pf?

packages/pn-pa-webapp/src/pages/components/AppStatus/useDownloadDocument.ts
clombardi: In fact no, otherwise I would have written it down. OK I do as you suggest ... even when I strongly disagree.

packages/pn-commons/src/components/EmptyState.tsx
AndreaCimini90: Yes your solution is good! We can avoid the use of IconType and leave KnownSentiments to maintain (as you said) the uniformity

packages/pn-pa-webapp/src/pages/components/AppStatus/useDownloadDocument.ts
AndreaCimini90: We use the download also in NotificationDetail and it seems to be the same of your hook.
Why you disagree with the proposal to make it generic?

packages/pn-pa-webapp/src/utils/role.utility.ts
AndreaCimini90: Mmmm at this point i should leave it as you have done

packages/pn-pa-webapp/src/pages/components/AppStatus/useDownloadDocument.ts
clombardi: Since I see the hook as too little and too dependent of the particular case (we should pass two props for a 6-line hook), is just a sensation. Maybe we could define some standard way to specify the info about a document to download, so that at least it would suffice with one prop. But as I told before, my ideas are not clear. And I tend to generalize only when I have a clear mindset. 
But I will proceed.

packages/pn-commons/src/components/EmptyState.tsx
clombardi: Refactor of `EmptyState` implemented, hope you like it!

packages/pn-pa-webapp/src/pages/components/AppStatus/useDownloadDocument.ts
clombardi: I moved `useDownloadDocument` into pn-commons, have to say it looks better than I thought.

AndreaCimini90 (approved):

PR #457, opened 2022-11-08
fix: pn-2487 adds multiple recipient with same notice code warning

LorenzoCocciarelli:
## Short description
In Recipient component, now if the user uses the same combination of creditorTaxId and 

- Adds check on same Notice Code and Creditor Tax Id for multiple recipients
- Refactor duplicate map and filter inside a function getDuplicateValuesByKeys
- Fixes tests for Recipient, now checks correctly that continue button is disabled if multiple recipients share the same Tax Id or Notice Code + Creditor Tax Id, and not because of other errors
- Changes some MockValues to ensure tests run correctly

## How to test
Send manual notification, in the recipients step, set the same combination of values in the "Codice fiscale ente creditore" and "Codice avviso" fields for 2 or more different recipients, an error should be displayed and the continue button should be disabled when all other fields are set correctly.

### DIFF
--- a/packages/pn-pa-webapp/public/locales/it/notifiche.json
+++ b/packages/pn-pa-webapp/public/locales/it/notifiche.json
@@ -172,7 +172,7 @@
         "title": "Destinatario",
         "fiscal-code-error": "Il codice fiscale inserito non è corretto",
         "identical-fiscal-codes-error": "C'è già un destinatario con questo Codice Fiscale. Inseriscine uno diverso.",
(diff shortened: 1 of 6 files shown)


AndreaCimini90 (changes requested):

LorenzoCocciarelli:
@AndreaCimini90 moved the function in notification.utility.ts and renamed variables with more generic names

ZorbaDimatteo (approved):

PR #478, opened 2022-11-21
Feat/pn 2676 limit other attachments to 10

LorenzoCocciarelli:
## Short description
Limits the number of other attachments to 10.

- The button is not visible when there are 11 attachment boxes (1 act + 10 other attachments)
- Adds a test to check that the button is not visible when there are 11 boxes

## How to test
In pn-pa-webapp send a new notification, in the attachments section add 10 attachments, the button to add another attachment should not be visible.

### DIFF
--- a/packages/pn-pa-webapp/public/locales/it/notifiche.json
+++ b/packages/pn-pa-webapp/public/locales/it/notifiche.json
@@ -204,6 +204,7 @@
       },
       "attachments": {
         "title": "Allegati",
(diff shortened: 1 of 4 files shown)


ZorbaDimatteo (changes requested):
@LorenzoCocciarelli let's put the max number of attachments as a constant

LorenzoCocciarelli:
@ZorbaDimatteo Added label to inform the user about the maximum number of attachments and constant that defines their number.

ZorbaDimatteo (approved):

PR #489, opened 2022-11-30
New notification - lost payment document infos after back action

AndreaCimini90:
## Short description
If new notification creation returns error and go back, the info about payment (ref and versionToken) are lost

- After upload documents, set ref and versionToken values in formik instance

## How to test
Go to the new notification. Use D’Arco as lastname of one recipient. Go forward to the last step and click the button "Invia". The creation of the notification goes in error due to the char ’. Go back to the second step (use back button and not the step button), change ’ with ', go to the last step and check that the notification is created with ref and versionToken, of the payment documents, not empty

### DIFF
--- a/packages/pn-pa-webapp/src/pages/components/NewNotification/PaymentMethods.tsx
+++ b/packages/pn-pa-webapp/src/pages/components/NewNotification/PaymentMethods.tsx
@@ -139,6 +139,10 @@ const PaymentMethods = ({ notification, onConfirm, isCompleted, onPreviousStep }
               hashHex: formikPagoPaForm.file.sha256.hashHex,
             },
           },
+          ref: {
+            key: formikPagoPaForm.ref.key,
+            versionToken: formikPagoPaForm.ref.versionToken,
+          },
         },
       };
       if (formikF24flatRate) {
@@ -153,6 +157,10 @@ const PaymentMethods = ({ notification, onConfirm, isCompleted, onPreviousStep }
               hashHex: formikF24flatRate.file.sha256.hashHex,
             },
           },
+          ref: {
+            key: formikF24flatRate.ref.key,
+            versionToken: formikF24flatRate.ref.versionToken,
+          },
         };
       }
       if (formikF24standard) {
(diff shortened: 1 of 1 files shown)


clombardi (changes requested):
Now the creation of a notification leads to an error, even for right values e.g. of surname. I just tried in the `main` branch, with exactly the same data, it went OK. Hence there should be some problem ... albeit I could not localize it.

AndreaCimini90:
I did some tests and the creation went well

clombardi (commented):
So sorry I answer only now, it actually took me most of the day to understand the bug solved by this PR.  
To my understanding, the changes actually solve the problem. But I think there is a possible simpler solution: to just eliminate the `setPaymentDocuments` action, and therefore also the `formatPaymentDocuments` function.

To support my opinion, I note that if the user goes back by using the stepper instead of the "Torna a Allegati" button, the problem about the ref doesn't show up, the reason being that the action is not executed (since `handlePreviousStep` is not invoked). In this scenario, the ref is indeed included in the payload of the notification creation.

I also tried the solution I'm suggesting implemented on the `main` branch, and it seems to work properly.

I also remark that there is still a problem regarding the handling of payment documents when a user retries to save a notification after an error. If the user **changes** the attachment and then retries, the second save will point to the same file as the precedent one, not to the newly attached file. The reason is that in such scenario, the `uploadNotificationPaymentDocument` action is not executed for the second file, since the `isCompleted` state is still set to `true`. I think it's the same problem reported in PN-2279, which in my opinion hasn't already been solved. 

The decision about proceeding to merge this PR depends, I guess, on the urgence to solve the particular problem which gave rise of it, noting that there is another, still unsolved, bug related to the same procedure.

clombardi (approved):
Now I kept understanding how the `ref`s are used, and I realize that they are needed to choose which documents to actually update when the action is called. Even if I guess that this is useless in the case of payment documents, because the action isn't called more than once (because of the `isCompleted` state), I guess is better to keep the logic coherent so that it's OK to merge the PR in its present form.

I still think that there is space to rethink how the information about the attachments and the "completed" signal are managed. The bug described in the previous comment is still there.

clombardi:
Still another issue: If I drop a document to upload (either in the attachment or in the payments page), and I go to a previous step (without going forward) using the stepper instead of the "Torna" button, then the document is not kept. That's because the `set...` action is not invoked.

AndreaCimini90:
For the navigation through stepper, there is another issue.
About isCompleted status, i fixed the bug and i ask you to test if it works

PR #500, opened 2022-12-06
PN 2865 - FE non registra gli errori api se errors vuoto

clombardi:
## Short description
This PR aims to fix a bug happening when an API call fails sending `data.errors = []` (so the `errors` attribute is present and the value is **exactly** the empty list). The FE didn't record correctly such fails, hence the toast indicating a problem did not appear, and the `ApiError` indication was not launched even if `ApiErrorWrapper` was used.

In principle, just a little adjustement in `pn-commons` which allows this case to be handled in the same way as if the `error` attribute weren't present.  
I profited to perform a little cleanup of unused code.

## How to test
You should have an API call that fails including `data.errors = []` in the response. At the moment of writing this description (2022.12.06 14.17 Italian time) you can test with the API Key option in pa-webapp, the API call which should retrieve the existing API keys fails in exactly this way. 
You can also simulate such an error. For the list of notifications in pa-webapp, just change the `getSentNotifications` function inside `NotificationsApi` with the following

```
  getSentNotifications: (params: GetNotificationsParams): Promise<GetNotificationsResponse> => {
    if (params) {
      return Promise.reject({ response: { status: 500, data: { errors: [] } } });
    }
    return Promise.reject({ response: { status: 501, data: { errors: [] } } });
  },
```

(the condicional is just to avoid getting errors for the unused `params`, i.e. to trick the compiler/linter).


ZorbaDimatteo:
@clombardi probably pn_ga and develop are aligned. Check with @AndreaCimini90  if it is possibile to change the target branch to develop (I thinks it's safe).

maurizioflauti (approved):

PR #503, opened 2022-12-07
pn-2855 - reset document ref data when it is uploaded by drag

AndreaCimini90:
## Short description
New notification - reset document ref data when it is uploaded and after is changed by drag

- Changed part that manage file upload in Attachments.tsx and PaymentMethods.tsx files

## How to test
Go to new notification, go to step 3 or 4, upload document and confirm. the file is uploaded (ref data is filled). Change the file by dragging new one without removing the old one. Reconfirm and check that the new file is uploaded and ref is filled with new data


clombardi (approved):
OK, the changes make sense to me, whenever a new file is uploaded, the ref is erased so that it is actually uploaded when the upload action is called.

PR #505, opened 2022-12-09
Pn 2763 fe multilingua sul portale landing

maurizioflauti:
## Short description
Update mui-italia to version 0.8.10.
Change pn-commons Footer to use language types from mui-italia LangSwitch component.
Add support for German (de), French (fr) and Slovenian (sl) languages on landing site.

- update mui-italia to v0.8.10
- refactor pn-commons Footer component to use types now exported by LangSwitch component
- add support to de, fr and sl LangCodes to landing site content manager
- add footer translation to landing site content manager
- add  en, de, fr and sl translations for both sections ("Cittadini" and "Enti") of the landing site

## How to test
Run the pn-landing application and verify you can choose any of the supported languages and see the page content changing accordingly.


packages/pn-landing-webapp/api/data/en.tsx
```diff
@@ -555,11 +538,11 @@ const paHorizontalNav: HorizontalNavProps = {
           <img src="static/icons/HORIZONTAL-NAV-1.svg" />
         </SvgIcon>
       ),
-      title: "Rappresenti un’impresa?",
+      title: "Do you represent a company?",
```
clombardi, line 541: I could not find any uses of neither `paHorizontalNav` nor `pfHorizontalNav`. I found the `getHorizontalNavData` in `api/index.ts`, but it seems to have no uses. Is this information included in view or a future version, or else it could be erased?

packages/pn-landing-webapp/src/layout/LandingLayout.tsx
```diff
@@ -53,45 +44,45 @@ const LandingLayout = ({ children }: Props) => {
...
-            onClick: () => window.open(pagoPALink.href, "_blank"),
+            ...appData.common.pagoPALink,
+            onClick: () => window.open(appData.common.pagoPALink.href, "_blank"),
           }}
-          legalInfo={companyLegalInfo}
-          postLoginLinks={postLoginLinks}
-          preLoginLinks={preLoginLinks()}
+          legalInfo={appData.common.companyLegalInfo}
+          postLoginLinks={appData.common.postLoginLinks}
+          preLoginLinks={appData.common.preLoginLinks}
```
clombardi, line 85: I guess that the "Prodotti e servizi" list of links can translated as well by providing a URL analogous to https://selfcare.pagopa.it/assets/products.json through the `productsJsonUrl` prop which is optional to `@pagopa/mui-italia/Footer` component. Unfortunately the very title "Prodotti e servizi" is fixed in the mui-italia code, some provision should be added for that.

clombardi (commented):
I'm not sure whether my comments are adequate or not, so I just leave the review to be further analyzed. I add a comment here which is not related to a particular point in the code.

I understand that the scope of this PR is limited to the "Enti" and "Cittadini" pages, excluding the documentation e.g. x.
But still in that page, there are some items which are not translated, and maybe they should. Namely: 
- the "Cittadini" and "Enti" links at the top.
- The text of the EU logo "Finanziato dall'Unione Europea", bottom-right.
- The timeline image in the PA page.

packages/pn-commons/src/utils/costants.tsx
```diff
@@ -1,8 +1,8 @@
-import { FooterLinksType, PreLoginFooterLinksType } from '@pagopa/mui-italia';
+import { FooterLinksType, Languages, PreLoginFooterLinksType } from '@pagopa/mui-italia';
 
 import { getLocalizedOrDefaultLabel } from '../services/localization.service';
 
-export const LANGUAGES = {
+export const LANGUAGES: Languages = {
```
clombardi, line 5: The language names are defined differently than in `pn-landing-webapp`. E.g. here it.en = "Inglese', while for `pn-landing-webapp` we have it.en="English".

clombardi (changes requested):
Change the decision to "Request changes" just to have "Changes requested" in the list of PRs.

packages/pn-landing-webapp/api/data/en.tsx
maurizioflauti: The HorizontalNav component has been removed from the layout but it may be a temporary decision. I received its translations so I think we'd better keep it and at most put a comment right above the getHorizontalNavData method.

packages/pn-landing-webapp/src/layout/LandingLayout.tsx
maurizioflauti: On this point I think the best approach would be to change the Footer component to allow, as already possible for any other link, these data to be passed in from the client code and then open a PR on mui-italia.
I wonder why they chose this solution, but perhaps there's a valid reason I'm missing about all this thing.

packages/pn-commons/src/utils/costants.tsx
maurizioflauti: In fact I put a comment where I defined the languages for the landing site under:pn-landing-webapp/src/layout/footer.constants.tsx because I think we need to choose one of the two possible solutions

maurizioflauti: I'm going to change the languages passed into the Footer of the landing site so that it will always show the content translated into the current selected language as decided during today's Standup.

packages/pn-landing-webapp/src/layout/LandingLayout.tsx
maurizioflauti: Following our discussion during today's Standup we need to better understand how to tackle this issue. Anyway I'm going to create a separate task on Jira.

clombardi (approved):
OK, great work foreseeing the eventual translated texts for the header. Ready to merge ...
... but still a slight detail (a "sciocchezza"): in the `Tab` elements inside `NavigationBar`, I wonder whether is better to set a fixed `key` and to translate only the `label`, e.g. 
```
key="Cittadini"   // fixed
label={pf}         // translated
```
instead of 
```
key={pf}         // translated
label={pf}         // translated
```

PR #527, opened 2022-12-22
error message for denomination #pn-3046

AndreaCimini90:
## Short description
Two different error message for denomination with length over 80 characters and with forbidden characters

- Changed validation in Recipient.tsx file

## How to test
- Login with pa -> go to new notification -> add denomination with length greater than 80 characters -> check both name and username are highlighted and that an error message is shown
- Login with pa -> go to new notification -> add a denomination with forbidden characters -> check that only the field with wrong char is highlighted and that an error message is shown


packages/pn-pa-webapp/public/locales/it/notifiche.json
```diff
@@ -178,6 +178,7 @@
         "pec-error": "Indirizzo PEC non valido",
         "too-long-denomination-error-PF": "La lunghezza totale di nome e cognome non può superare gli 80 caratteri",
```
ZorbaDimatteo, line 179: @AndreaCimini90 this message should be "Inserisci max 80 caratter" (also for pg)

ZorbaDimatteo: we should use the same message for pf and pg

ZorbaDimatteo (changes requested):
@AndreaCimini90 merge is blocked since this branch is not updated with target branch

ZorbaDimatteo (approved):

PR #530, opened 2022-12-27
PN-2913-fe-pf-webapp-cleans-token-from-hash

clombardi:
## Short description
The `#token` hash entry in the URL is cleaned on every access to personafisica-webapp from a logged user, whenever such entry is present.  
This avoids the token to be part to the "origin" sent to personafisica-login in case of a timeout-caused logout, and therefore to have eventually two tokens in the personafisica-webapp URL.

Just changed `SessionGuard` in personafisica-webapp to remove the `#token` element from the hash in the location object which represents the URL, for accesses with a logged user (e.g. excluding anonymous accesses). The remaining hash elements, as well as the search string, are preserved.  
Note that given the need to analyze and eventually change the location, now all the URL invocations mediated by the SessionGuard perform a navigation.  

This PR implies **no changes** for pa-webapp, since for this application the login procedure (in this case performed through selfcare) does not preserve any part of the URL; there is no such thing as an "origin" which is passed to selfcare.

## How to test
In a browser without any current session, enter the following in the address bar:
> \<personafisica-webapp-URL\>/deleghe
Go through the login procedure provided by SPID, you will login into the personafisica webapp and see the delegates/delegators ("Deleghe") page. As a first indication, you should not see the token in the browser' address bar.
image

Wait for the automatic logout due to timeout, the choice of login method should appear automatically
image

Perform the login procedure, you should login again into the personafisica webapp normally, seeing again the "Deleghe" page.


LorenzoCocciarelli (approved):
It's fine for me

PR #531, opened 2022-12-28
PN-3065-fe-invio-manuale-visualizzazione-errori-navigazione

clombardi:
## Short description
Consider this scenario: In the manual registration of a new notification ("inserimento manuale notiifca"), the user performs the following sequence:
1. fill the fist step ("Informazioni preliminari") so that she can proceed to the second step ("Destinatario").
2. In the second step, fill a field with a wrong value (e.g. "Codice avviso" with anything different than a 18-numeric-char string).
3. Go back to the first step, through either the "Torna a informazioni preliminari" button bottom-left, or the corresponding link in the stepper above the form.
4. Proceed again to the second step.

Without this PR, the user doesn't see the error messages for the wrong-filled fields.
image

The aim of this PR is to show the messages in the described scenario.
image

Just changed the ad-hoc `FormTextField` to make the error message to appear whenever the value of the field, even when the field has not been "touched".  

**NB**  
Recall that the `Formik` form is re-created each time the user gets into a step. This issue could have been avoided if we would handle such forms outside the particular step components, i.e. in the `NewNotification` component, and pass them to each step component as a prop ... as we already discussed inside the FE team.

## How to test
Reproduce the scenario described above.


LorenzoCocciarelli (approved):
It's ok for me

PR #535, opened 2023-01-02
refactoring: delete errorDeleghe and label used by it

leleOFA:
## Short description
delete ErrorDeleghe component because is used but never see it

- Delete ErrorDeleghe component in NuovaDelega page
- delete label never used

## How to test
with an error see another component that render error in esplicit message


clombardi (approved):
All OK, the special error cases are managed by `PFAppErrorFactory`, this factory includes the current error codes sent by the BE.

PR #540, opened 2023-01-10
Added tax code validation in filter notifications PA

NicolaGiornetta:
## Short description
Added in the notification filter in PA tax code validation togheter with fiscal code.

- Validation for recipient id works with both tax code and fiscal code regular expressions

## How to test
Tax code is 11 digits long. Just type a number with that amount of digits in the filter notifications in PA. Then you should be able to click "Apply filters".

### DIFF
--- a/packages/pn-pa-webapp/src/pages/components/Notifications/FilterNotifications.tsx
+++ b/packages/pn-pa-webapp/src/pages/components/Notifications/FilterNotifications.tsx
@@ -67,6 +67,8 @@ function isFilterapplied(filtersCount: number): boolean {
   return filtersCount > 0;
 }
 
+const getValidStatus = (status: string) => (status === 'All' ? '' : status);
+
 const FilterNotifications = forwardRef(({ showFilters }: Props, ref) => {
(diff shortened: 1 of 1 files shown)


packages/pn-pa-webapp/src/pages/components/Notifications/FilterNotifications.tsx
```diff
@@ -78,7 +80,10 @@ const FilterNotifications = forwardRef(({ showFilters }: Props, ref) => {
   const validationSchema = yup.object({
     recipientId: yup
       .string()
-      .matches(dataRegex.fiscalCode, t('filters.errors.fiscal-code', { ns: 'notifiche' })),
+      .matches(
+        new RegExp(dataRegex.fiscalCode.source + '|' + dataRegex.pIva.source),
```
LorenzoCocciarelli, line 84: @NicolaGiornetta testando la regex ho visto che prima non era case sensitive mentre ora lo è, potremmo aggiungere la flag 'i' come parametro per farla tornare case insensitive?

LorenzoCocciarelli (approved):

PR #542, opened 2023-01-11
Removed validation check for payment methods

NicolaGiornetta:
## Short description
Removed validation check for payment methods.

- Step 4, Payment methods file upload are no longer required
 
## How to test
Create a new notification with at least one payment method in Step 1; fill Step 2 and 3 as usual; Step 4 leave empty. You should be able to click "create" button anyway.


clombardi:
Mi sono preso io questa PR per fare la review.

packages/pn-pa-webapp/src/pages/components/NewNotification/PaymentMethods.tsx
```diff
@@ -249,13 +210,36 @@ const PaymentMethods = ({
...
+  };
+
   const formik = useFormik({
     initialValues,
-    validationSchema: notification.paymentMode !== PaymentModel.NOTHING ? validationSchema : null,
     validateOnMount: true,
     onSubmit: async (values) => {
-      if (isCompleted) {
-        onConfirm();
+      const emptyForm = formIsEmpty(values);
```
clombardi, line 235: I was quite confused while trying to figure out the effect of the modifications. Maybe is clearer if the first conditional statement is left as it was, and we just change the condition of the second one to 
```
if (notificationPaymentMode !== PaymentModel.NOTHING && !emptyForm)
```
or perhaps even
```
if (!emptyForm)
```
adding a comment to notice that `(notificationPaymentMode === PaymentModel.NOTHING` implies `emptyForm`.

What do you think about this?

clombardi (changes requested):
Liberate the comments, hope Nicola is able to see them.

clombardi (approved):
This was indeed a **difficult** one. I tried with everything I could think about, seems to work properly in all scenarios. Happily approved.

**NB**:
we are aware about the fact a Cypress test is failing. Is a test about the notification display, is not connected with this PR.

PR #561, opened 2023-01-23
Chore: transform taxId in uppercase before notification creation

NicolaGiornetta:
## Short description
transform taxId in uppercase before notification creation.

- Added taxId transformation in newNotificationRecipientsMapper() function

## How to test
Create a new notification and insert tax id in lowercase then check the payload of API notification creation. You'll see the taxId in uppercase


packages/pn-pa-webapp/src/utils/notification.utility.ts
```diff
@@ -49,7 +49,7 @@ const newNotificationRecipientsMapper = (
     return {
       denomination: recipient.recipientType === RecipientType.PG ? recipient.firstName : `${recipient.firstName} ${recipient.lastName}`,
       recipientType: recipient.recipientType,
-      taxId: recipient.taxId,
+      taxId: recipient.taxId.toUpperCase(),
```
AndreaCimini90, line 52: This is causes problems, because the paymentDocuments (below in the code) is an object where the keys are the recipients tax ids.
I suggest to do the transformation into the moethod saveRecipients in reducers.ts file

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #563, opened 2023-01-24
PN-3842 fix: adjusts clickable areas for checkboxes in Recipient

LorenzoCocciarelli:
## Short description
Previously the addresses checkbox clickable areas overlapped in between them, clicking on the lower area of the top one activated the bottom one, after this changes they should not overlap anymore and the clickable zone of each one should be more clear.

-  Adjusts clickable areas for checkboxes in Recipient

## How to test
Click on the lower area of the link 'Aggiungi un domicilio digitale', it should add the relative text field correctly.


packages/pn-pa-webapp/src/pages/components/NewNotification/Recipient.tsx
```diff
@@ -562,7 +562,7 @@ const Recipient = ({
                         />
                       )}
                     </Grid>
-                    <Grid container spacing={2}>
```
AndreaCimini90, line 565: This change breaks physical address spacing

AndreaCimini90 (changes requested):

packages/pn-pa-webapp/src/pages/components/NewNotification/Recipient.tsx
LorenzoCocciarelli: Restored physical address spacing

PR #566, opened 2023-01-25
feat/PN-3478 regex-telefono-email

clombardi:
## Short description
Defined a specific regex for email address validation, use it instad of `yup.email()`. Made it compatible (in fact more strict) than the one used BE-side. 

Change the regex defined for phone numbers, now it does not accept the space between the first three digits and the rest.  
**NB**: accepts only Italian mobile phone numbers, including the restriction about starting with "3".

- Used new regexes for email and phone thorough Contacts page in PF/PG
  - Legal PEC email address, insert and update.
  - Courtesy email address, insert and update.
  - Courtesy phone number for SMS, insert and update.
  - Additional contacts (which can be either PEC, courtesy email or SMS phone number): insert and update.
- Used new regex for email in manual notification creation FE, recipients step, digital domicile which is an email address.

## How to test
Just play with the many options in Contacts page and the recipients step in manual notification creation, setting both valid and invalid value.  

**Beware**  
In the update of any phone number in Contacts page, you must delete the "+39" prefix, and afterwards if you try to save the modification you will get an error. This is a known bug which is being handled by a separate PR.


clombardi:
This PR would be complete and ready to be reviewed ... but ... the Cypress tests yield several errors. I keep it "parked" until I understand the right thing to do.

LorenzoCocciarelli (approved):

PR #584, opened 2023-02-07
Big refactor CDN OneTrust handling

NicolaGiornetta:
## Short description
Refactored CDN OneTrust handling by using env variables

## List of changes proposed in this pull request
- Added compileOneTrustPath function in pn-commons utilities subpath
- Added new ENV variables

## How to test
First off you need to add new variables in your .env.development.local. See below:

## pn-landing-webapp
NEXT_PUBLIC_ONE_TRUST_DRAFT_MODE = ''
NEXT_PUBLIC_ONE_TRUST_PP = b5c8e1dc-89df-4ec2-a02d-1c0f55fac052

## pn-pa-webapp
REACT_APP_ONE_TRUST_DRAFT_MODE = draft
REACT_APP_ONE_TRUST_PP = 4824a110-316e-42fd-b492-8e2b0513db70
REACT_APP_ONE_TRUST_TOS = 083fb982-149c-4241-be09-12ae67d88b66

## pn-personafisica-login
REACT_APP_ONE_TRUST_DRAFT_MODE = draft
REACT_APP_ONE_TRUST_PP = 133242e7-ad9a-45e5-a417-b7a50c746899

## pn-personafisica-webapp
REACT_APP_ONE_TRUST_DRAFT_MODE = draft
REACT_APP_ONE_TRUST_PP = 365c84c5-9329-4ec5-89f5-e53572eda132
REACT_APP_ONE_TRUST_TOS = b0da531e-8370-4373-8bd2-61ddc89e7fa6
REACT_APP_ONE_TRUST_PARTICIPATING_ENTITIES = ffb2a640-8165-4d5f-94c2-6259e21bee51

## pn-personagiuridica-webapp
REACT_APP_ONE_TRUST_DRAFT_MODE = draft
REACT_APP_ONE_TRUST_PP = 365c84c5-9329-4ec5-89f5-e53572eda132
REACT_APP_ONE_TRUST_TOS = b0da531e-8370-4373-8bd2-61ddc89e7fa6

Then open Privacy Policy, Terms of Service, Participating Entities in webapps.

PR #585, opened 2023-02-08
Test/pn 3402 fe e 2 e test pipeline per progetti fe

maurizioflauti:
## Short description
Add support to execute every cypress test available inside pa and pf webapps.

- add cypress_buildspec.yaml
- add .env.cypress and start-waiton dependency for both pf-webapp and pa-webapp


cypress_buildspec.yaml
```diff
@@ -0,0 +1,44 @@
+version: 0.2
+env:
+  shell: bash
+  secrets-manager:
+    SONAR_TOKEN: arn:aws:secretsmanager:eu-central-1:911845998067:secret:SonarCloud-7370vM:token
```
ZorbaDimatteo, line 5: @maurizioflauti i think this token is not used in this builder, delete it

packages/pn-personafisica-webapp/.env.cypress
```diff
@@ -0,0 +1,16 @@
...
+REACT_APP_CHECKOUT_URL='https://uat.checkout.pagopa.it'
+REACT_APP_PAYMENT_DISCLAIMER_URL='https://www.pagopa.it'
+HOST='portale.dev.pn.pagopa.it'
+WEBAPP_URL='https://portale.dev.pn.pagopa.it:443'
+REACT_APP_MIXPANEL_TOKEN='DUMMY'
+REACT_APP_ONETRUST_DOMAIN_ID='fd5aef6f-a6d3-422b-87b7-aa5e2cb6510c'
+REACT_APP_PAGOPA_HELP_EMAIL='notifichedigitali@assistenza.pagopa.it'
+REACT_APP_VERSION=$npm_package_version
+REACT_APP_DISABLE_INACTIVITY_HANDLER=true
+REACT_APP_ONE_TRUST_PORTAL_CDN_PP='https://privacyportalde-cdn.onetrust.com/77f17844-04c3-4969-a11d-462ee77acbe1/privacy-notices/draft/365c84c5-9329-4ec5-89f5-e53572eda132.json'
```
ZorbaDimatteo, line 12: @maurizioflauti a lot of this env variables are going to change with this pr https://github.com/pagopa/pn-frontend/pull/584

ZorbaDimatteo (commented):
@maurizioflauti as long as wait-on suits well our goals for this task, I'd prefer to avoid add another dependency in our toolchain to mantain.
I would suggest to use directly node setTimeout funcionality creating a script similar to that
"timeout": "node -e \"require('node:timers/promises').setTimeout(5_000)\"",
This script should tun smoothly on every OS

what do you think?

cypress_buildspec.yaml
```diff
@@ -0,0 +1,44 @@
...
+  pre_build:
+    commands:
+      - echo launching yarn install in $SUB_PRJ_NAME...
+      - cd $SUB_PRJ_NAME
+      - yarn install --frozen-lockfile
+  build:
+    commands:
+      - BRANCH_NAME=$(echo $CODEBUILD_WEBHOOK_HEAD_REF | sed 's/refs\/heads\///'); export BRANCH_NAME=${BRANCH_NAME:=$GIT_DEFAULT_BRANCH}
+      - if echo $CODEBUILD_SOURCE_VERSION | grep -q ^pr/ ; then PR_NUM=$(echo $CODEBUILD_SOURCE_VERSION | sed 's/pr\///'); BRANCH_TARGET=$(echo $CODEBUILD_WEBHOOK_BASE_REF | sed 's/refs\/heads\///'); export PR_NUM BRANCH_TARGET; fi;
+      ## Running the development server
```
ZorbaDimatteo, line 29: @maurizioflauti from line 27 to line 29 we set and export some env variables but we don't use them within cypress tasks. I think we should remove these lines

cypress_buildspec.yaml
maurizioflauti: ok

packages/pn-personafisica-webapp/.env.cypress
maurizioflauti: ok

cypress_buildspec.yaml
maurizioflauti: ok

maurizioflauti: @ZorbaDimatteo using setTimeout instead of wait-on would imply having to "manually" tune the timer instead of "dynamically" wait for the server to be up and running. This could lead to either have a more likely to fail pipeline (if, for some reason, the server takes longer than we expect and we went for a restrictive timer) or to waste a lot of time keep waiting, without any good reason, while the server is already up and running (if we define a conservative timeout). In this case I would have no doubts that the advantages of having a service tapping on my shoulders when the startup is done are well worth the downsides of installing a new (dev)dependency.

maurizioflauti:
@ZorbaDimatteo using setTimeout instead of wait-on would imply having to "manually" tune the timer instead of "dynamically" wait for the server to be up and running. This could lead to either have a more likely to fail pipeline (if, for some reason, the server takes longer than we expect and we went for a restrictive timer) or to waste a lot of time keep waiting, without any good reason, while the server is already up and running (if we define a conservative timeout). In this case I would have no doubts that the advantages of having a service tapping on my shoulders when the startup is done are well worth the downsides of installing a new (dev)dependency.

ZorbaDimatteo (approved):

PR #588, opened 2023-02-10
Pn 3866 fe gestire gruppi in fase di invio manuale della notifica

NicolaGiornetta:
## Short description
Added hasGroups parameter to auth.api of PA and modified validation of Preliminary Informations

Accordi to the task, the preliminary informations have now new validation for groups. If hasGroups in userData is present and true, the field is set as required, even if there are groups available.

## How to test
Two tests:
- You have to find a user who have hasGroups set to true and then preliminary informations will required the field to continue
- You have to find a user who have hasGroups set to false and then preliminary informations will NOT require the field to continue


packages/pn-pa-webapp/src/pages/components/NewNotification/PreliminaryInformations.tsx
```diff
@@ -54,7 +55,7 @@ const PreliminaryInformations = ({ notification, onConfirm }: Props) => {
     subject: yup.string().required(`${t('subject')} ${tc('required')}`),
     physicalCommunicationType: yup.string().required(),
     paymentMode: yup.string().required(),
-    group: groups.length > 0 ? yup.string().required() : yup.string(),
+    group: (groups.length > 0 && user.hasGroups) ? yup.string().required() : yup.string(),
```
LorenzoCocciarelli, line 58: we could avoid checking for groups length using hasGroups, if hasGroups is false groups.length would alsa be 0

packages/pn-pa-webapp/src/api/auth/Auth.api.ts
```diff
@@ -13,6 +13,7 @@ export const AuthApi = {
             fiscal_number: response.data.fiscal_number,
             organization: response.data.organization,
             sessionToken: response.data.sessionToken,
-            uid: response.data.uid
+            uid: response.data.uid,
+            hasGroups: response.data.hasGroups,
```
LorenzoCocciarelli, line 17: hasGroups should be a field inside organization

packages/pn-pa-webapp/src/redux/auth/reducers.ts
```diff
@@ -29,6 +29,7 @@ const organizationMatcher = yup.object({
 const userDataMatcher = yup
   .object({
     ...basicUserDataMatcherContents,
+    hasGroups: yup.boolean(),
```
LorenzoCocciarelli, line 32: hasGroups should be inside the organizationMatcher

LorenzoCocciarelli (changes requested):

packages/pn-pa-webapp/src/api/auth/Auth.api.ts
NicolaGiornetta: so, I should have response.date.organization.hasGroup instead?

LorenzoCocciarelli: yes, it should

LorenzoCocciarelli (approved):

PR #589, opened 2023-02-10
Feat/PN-3623 how to show digital send failure in timeline

clombardi:
## Short description
Verified how the timeline of a notification is shown in the scenario where the digital send procedure fails (eventually after many retries) and a simple-registered ("raccomandata semplice") letter is send to complete the shipment of the notification.

- Shifted some timeline events from the DELIVERED state to the DELIVERING state. Namely, up to SEND_SIMPLE_REGISTERED_LETTER or DIGITAL_FAILURE_WORKFLOW (the latter if both are present).
- Changed the copy of the DELIVERED state to show whether the notification was sent to a digital address or to an analog one.
- Changed the copy of the SEND_DIGITAL_FEEDBACK timeline element, for both the OK and KO states, in favor of a more explicative description.

## How to test
Access to the timeline inside the notification detail page in two scenarios.
1. The digital shipment succeeds, maybe after some retries. <br/> Example (in DEV environment): IUN LPYJ-QPGN-RATM-202302-L-1, recipient Lucrezia Borgia, sender Comune di Milano.
2. The digital shipment fails and the shipment is completed through the sending of a "raccomandata semplice". <br/> Example (in DEV environment): IUN JYTK-QPVE-KQWK-202302-A-1, recipient Lucrezia Borgia, sender Comune di Palermo.


packages/pn-personafisica-webapp/public/locales/it/notifiche.json
```diff
@@ -105,10 +105,10 @@
       "send-digital-domicile-error-description": "L'invio della notifica a {{name}} all'indirizzo PEC {{address}} non è riuscito.",
       "send-digital-domicile-success": "Invio via PEC riuscito",
       "send-digital-domicile-success-description": "L'invio della notifica a {{name}} all'indirizzo PEC {{address}} è riuscito.",
-      "send-digital-error": "Invio per via digitale non riuscito",
-      "send-digital-error-description": "Il tentativo di invio della notifica per via digitale a {{name}} non è riuscito.",
-      "send-digital-success": "Invio per via digitale riuscito",
-      "send-digital-success-description": "Il tentativo di invio della notifica per via digitale a {{name}} è riuscito.",
+      "send-digital-error": "Invio via PEC non riuscito",
```
AndreaCimini90, line 108: Same changes must be carried in pg and pa

packages/pn-commons/src/utils/notification.utility.ts
```diff
@@ -436,7 +500,7 @@ function populateMacroSteps(parsedNotification: NotificationDetail) {
     for (const timelineElement of status.relatedTimelineElements) {
       populateMacroStep(parsedNotification, timelineElement, status, acceptedStatusItems);
     }
-    // order step by time
+    // order step by time, latest first
```
AndreaCimini90, line 503: Maybe we can do the shift inside this loop and before reorder?!

packages/pn-commons/src/utils/notification.utility.ts
```diff
@@ -51,14 +53,19 @@ function localizeStatus(
...
+
 /**
  * Returns the mapping between current notification status and its color, label and descriptive message.
  * @param  {NotificationStatus} status
  * @returns object
  */
 export function getNotificationStatusInfos(
   status: NotificationStatus,
-  recipient?: string
+  options?: { recipient?: string; completeStatusHistory?: Array<NotificationStatusHistory> }
```
AndreaCimini90, line 68: In general I should avoid logic inside this function. There is a mapper function that do all the transformations before sending data to the component.
I should set a custom generic property in the NotificationStatusHistory for example:

```
export interface NotificationStatusHistory {
  status: NotificationStatus;
  activeFrom: string;
  relatedTimelineElements: Array<string>;
  // only fe
  steps?: Array<INotificationDetailTimeline>;
  details?: BaseDetails | DeliveredDetails;
}

interface BaseDetails {
 recipient?: string
}

interface DeliveredDetails extends BaseDetails {
  deliveringMode?: {enumeration for delivering modes}
}
```

and I populate this genric property in the mapper. In this way we'll put all transformations in one method and we'll avoid extra loops over notification object

clombardi: Even when I will an alternative solution following (I hope!) the lines you suggest, I strongly disagree with your view, since it concentrates the complexity in a single function (i.e. the mapper function) and data structure (NotificationStatusHistory) that becomes more and more difficult to understand as we add more details into it. 
In this particular case, we are adding an attribute in the general NotificationStatusHistory object, that makes sense only for the DELIVERED instance.
On the other hand, in my opinion performance gains that change only the quantity of operations but not the corresponding order (linear / logarithmic / polynomial / exponential) are of no practical benefit in the vast majority of cases, while isolating the individual tasks or aspects in different functions simplify the code where the gain in readibility is greater than the loss of (temporal) performance.
Indeed, I would rather go the other way around, namely in the direction of deflating populateMacroSteps and define separated functions for the diffent tasks done there, i.e. duplicating the ACCEPTED-related timeline elements hidden in one state and with no legal facts in the other one, setting recipient to the NotificationStatus.VIEWED state, and changing the VIEWED status to VIEWED_AFTER_DEADLINE if the notification passed through EFFECTIVE_DATE before.

In general, given a situation in which I can either pre-compute a datum and store it inside a data structure, or compute it in the place in which it is used, I tend to choose the second alternative, because it leads to more cohesive, less coupled and less complicated code. I taught this idea for 20 years and I'm still convinced of it.

packages/pn-commons/src/utils/notification.utility.ts
clombardi: Done, even when I don't like the new code. In my opinion, the `populateMacroSteps` became quite unreadable. I inserted all the comments that came out of my mind.

packages/pn-commons/src/utils/notification.utility.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable sonarjs/cognitive-complexity */
```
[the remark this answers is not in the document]
clombardi: I refuse to do this. There are two functions that (in Sonar' opinion) have a too great complexity
- `getLegalFactLabel` which I didn't touch at all in this PR.
- `populateMacroSteps`, which complexity jumped to 46 because ... of the modifications you suggest in your other comments.

packages/pn-commons/src/utils/notification.utility.ts
AndreaCimini90: This is a transitional phase, in the near future the mapper function will be mooved to a lambda function in back-end.
So to have all the transformations in one place, it will help us to do this transition

AndreaCimini90 (approved):

AndreaCimini90 (approved):

PR #598, opened 2023-02-15
fix/pn-3896

maurizioflauti:
## Short description
This pr should fix PP, ToS and ParticipatingEntities pages tests failure in ci when REACT_APP_ONE_TRUST_DRAFT_MODE environment variable is set, for some reason (buildspec.yaml), to true.


clombardi (approved):
I agree with the changes as proposed, OK to merge.

PR #605, opened 2023-02-20
Fixed copy about login button in PA

NicolaGiornetta:
## Short description
Fixed copy about login button in PA.

- Modified common.json in PA

## How to test
The button shown in PA before login should be "Accedi" now instead of "Procedi al login"


clombardi (approved):
OK to merge

PR #608, opened 2023-02-21
Chore: Added indexed key component to avoid duplication

NicolaGiornetta:
## Short description
Added indexed key component to avoid duplication.

- Added indexed key component to avoid duplication

## How to test
Open a multi addresse notification and check the web browser console. Now the errors about React key duplication are gone.

### DIFF
--- a/packages/pn-commons/src/components/NotificationDetail/NotificationDetailDocuments.tsx
+++ b/packages/pn-commons/src/components/NotificationDetail/NotificationDetailDocuments.tsx
@@ -34,7 +34,7 @@ const NotificationDetailDocuments = ({
 
   const mapOtherDocuments = (documents: Array<NotificationDetailDocument>) => documents.map((d) => {
     const document = {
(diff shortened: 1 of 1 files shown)


packages/pn-commons/src/components/NotificationDetail/NotificationDetailDocuments.tsx
```diff
@@ -32,13 +32,13 @@ const NotificationDetailDocuments = ({
   // TODO: remove comment when link ready downloadFilesLink
 }: Props) => {
 
-  const mapOtherDocuments = (documents: Array<NotificationDetailDocument>) => documents.map((d) => {
+  const mapOtherDocuments = (documents: Array<NotificationDetailDocument>) => documents.map((d, index) => {
```
AndreaCimini90, line 35: The use of the index is not recommended. Into the object from the be and in the timeline, there is an id or something similar?

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #611, opened 2023-02-23
Pn 4264 fe integration test su invio manuale senza pagamenti cypress

NicolaGiornetta:
## Short description
Added Cypress ENV variable to switch between notification with / without payment methods.

- New boolean variable for cypress.env.json IS_PAYMENT_ENABLED. This variable must be set the same value as REACT_APP_IS_PAYMENT_ENABLED

## How to test
Two ways of testing available.
1. Set REACT_APP_IS_PAYMENT_ENABLED and IS_PAYMENT_ENABLED to true to enable notifications with payment methods
2. Set REACT_APP_IS_PAYMENT_ENABLED and IS_PAYMENT_ENABLED to false to disable notifications with payment methods

## Update Confluence configuration page
When PR is approved then update the Confluence configuration page, section Cypress for PA, by adding the new variable IS_PAYMENT_ENABLED in cypress.env.json

### DIFF
--- a/packages/pn-pa-webapp/cypress/e2e/All.cy.ts
+++ b/packages/pn-pa-webapp/cypress/e2e/All.cy.ts
@@ -2,5 +2,6 @@ import "./Auth.cy";
 import "./TosAgreement.cy";
 import "./FilterNotifications.cy";
 import "./NewNotification.cy";
+import "./NewNotification.without.payment.methods.cy";
 import "./NotificationDetail.cy";
 import "./Pagination.cy";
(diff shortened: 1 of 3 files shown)


AndreaCimini90 (commented):
I'm not sure about the usage of the cypress var, because it is tied up to the application var.
I'd prefer to skip the paid tests as there will be a separate env when we'll create the  cypress aws pipeline.
what do you think @ZorbaDimatteo?

ZorbaDimatteo:
@AndreaCimini90 I agree with you, please @NicolaGiornetta can you refactor these tests considering this comment?

NicolaGiornetta:
@AndreaCimini90 , @ZorbaDimatteo 
Done! I removed the variable from Cypress env and its usage. Set to skip() the tests with payment methods.

AndreaCimini90 (approved):

PR #619, opened 2023-03-01
new notification - wrong tax id error - pn-4421

AndreaCimini90:
## Short description
Changed error message when an user send a new notification request with worng tax id

- Added error factory
- Added a class to manage invalid parameter error
- Added localization for invalid parameter error
- Added tests

## How to test
Go to new notification -> add a recipient with TMOCHR95S51H501H or any wrong tax id -> check that right error message is shown

### DIFF
--- a/packages/pn-commons/src/utils/AppError/AppError.ts
+++ b/packages/pn-commons/src/utils/AppError/AppError.ts
@@ -1,21 +1,20 @@
-import {
-  ServerResponseError,
-  ErrorMessage,
-  AppResponseError } from '../../types/AppResponse';
+import { ServerResponseError, ErrorMessage, AppResponseError } from '../../types/AppResponse';
 
 abstract class AppError {
   protected code: string;
   protected element: string;
+  protected detail: string;
 
   constructor(error: ServerResponseError) {
     this.code = error.code;
-    this.element = error.element || "";
+    this.element = error.element || '';
+    this.detail = error.detail || '';
   }
 
   getErrorDetail(): ServerResponseError {
     return {
       code: this.code,
-      element: this.element
+      element: this.element,
     };
   }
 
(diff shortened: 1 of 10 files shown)


packages/pn-pa-webapp/src/utils/AppError/PAAppErrorFactory.ts
```diff
@@ -0,0 +1,29 @@
...
+  AppErrorFactory,
+  ServerResponseError,
+  UnknownAppError,
+} from '@pagopa-pn/pn-commons';
+
+import { GenericInvalidParameterAppError } from './GenericInvalidParameterAppError';
+import { ServerResponseErrorCode } from './types';
+
+export class PAAppErrorFactory extends AppErrorFactory {
+  private translateFunction: (path: string, ns: string) => string = (path: string) => path;
```
clombardi, line 12: why the default translate function? If it's for test purposes, please add a comment about it. Otherwise, I think I would prefer the app to break if no translation function has been provided to the error factory.

packages/pn-commons/src/utils/AppError/AppError.ts
```diff
@@ -1,21 +1,20 @@
...
-    this.element = error.element || "";
+    this.element = error.element || '';
+    this.detail = error.detail || '';
   }
 
   getErrorDetail(): ServerResponseError {
     return {
       code: this.code,
-      element: this.element
+      element: this.element,
```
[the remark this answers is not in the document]
AndreaCimini90: You are right. I don't know why there are these methods

packages/pn-pa-webapp/src/utils/AppError/PAAppErrorFactory.ts
AndreaCimini90: It's not for test purpose, but if you check the constructor you see that we are forced to pass a translate function. So the app cannot break.
On the other side the valorization inside the class is useless and maybe the definition it's enough

clombardi (approved):
All clean new, OK to merge.

PR #630, opened 2023-03-07
fix(pn-4569): documents availability in pa-webapp

LorenzoCocciarelli:
## Short description
Fixes logic for hasDocumentsAvailable in pa-webapp

- fixes logic for hasDocumentsAvailable in pa-webapp

## How to test
Documents should be available for not cancelled and viewed notification since less than 120 days


clombardi (approved):
Sound modification, OK to merge.

PR #633, opened 2023-03-07
fix: add tests for appStatus reducers, events and external-registries

LorenzoCocciarelli:
## Short description
Add tests for appStatus reducers, events and external-registries api for pf and pg apps

## How to test
Run unit tests in pf and pg webapps


NicolaGiornetta (commented):
Tests are ok, including coverage. However there are some where it outputs some errors, even if they passes. Here the list:

PF
NotificationPayment.test.tsx
SpecialContactElem.test.tsx

PG 
App.test.tsx
NotificationPayment.test.tsx
SpecialContactElem.test.tsx

About the external-registries API tests (both PF and PG): should we also add a test for rejected status?

NicolaGiornetta (approved):

PR #644, opened 2023-03-13
Replaced translations for landing app, downgraded MUI Italia package

NicolaGiornetta:
## Short description
Replaced translations for landing app.

- Replaced translations for SL, EN, FR and DE languages in landing app
- @pagopa/mui-italia package has been downgraded from 0.9.0 to 0.8.12 as temporary fix for not launching app

## How to test
You need to reinstall packages in pn-landing before testing.
Then switch languages to see the new translations.


packages/pn-landing-webapp/api/data/en.tsx
```diff
@@ -48,10 +48,10 @@ const onReadClick = () => {
 };
 
 const navigation: INavigationBarProps = {
-  title: "Piattaforma Notifiche",
+  title: "Notifications Platform",
```
clombardi, line 51: This title is translated into English and Slovenian, but not in French neither in German, is that on purpose?

clombardi (commented):
I'm not sure whether is right to translate the title "Piattaforma Notifiche" into two languages and not into the other two. I would prefer to clarify this detail before approving the PR.

packages/pn-landing-webapp/api/data/en.tsx
NicolaGiornetta: Good question and I can't answer. Bring this to the SU.

clombardi: We discussed this issue in the SU, the decision was to merge this PR as is and expose the details about translations in future meetings with other staff linked to PN.

clombardi (approved):
Works as defined (though not as expected, see resolved conversation), OK to merge.

PR #648, opened 2023-03-15
Replacing old Piattaforme Notifiche strings with SEND

NicolaGiornetta:
## Short description
Replaced Piattaforme Notifiche strings with SEND.

- Replaced Piattaforme Notifiche strings with SEND

This affects to OneTrust JSONs too and translations in landing-app. I suggest an in-deep verification for these.


LorenzoCocciarelli:
In onetrust files, every occurrence of "SEND" is followed by the word "Digitali", I think it should be removed.

PR #665, opened 2023-03-29
Fix a11y button in EmptyState component

NicolaGiornetta:
## Short description
Fix a11y button in EmptyState component.

- Added button with a11y via keyboard to EmptyState component

## How to test
You should be able to navigate on button "Add new delegate" now in page Your delegates in PF.

### DIFF
--- a/packages/pn-commons/src/components/EmptyState.tsx
+++ b/packages/pn-commons/src/components/EmptyState.tsx
@@ -1,9 +1,10 @@
 import { ReactNode } from 'react';
 import { Typography, Box } from '@mui/material';
 import { SvgIconComponent } from '@mui/icons-material';
+import { ButtonNaked } from '@pagopa/mui-italia';
 import { KnownSentiment } from '../types';
 import { iconForKnownSentiment } from '../types/EmptyState';
-
+import { buttonNakedInheritStyle } from '../utils';
 type Props = {
   /** Callback to be called when performing an empty action */
   emptyActionCallback?: (e: any, source?: string) => void;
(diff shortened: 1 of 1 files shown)


packages/pn-commons/src/components/EmptyState.tsx
```diff
@@ -47,29 +49,25 @@ function EmptyState({
...
-            variant="body2"
-            fontWeight={'bold'}
+          <ButtonNaked
             data-testid="callToActionFirst"
-            sx={{
-              cursor: 'pointer',
-              display: 'inline',
-            }}
             onClick={emptyActionCallback}
+            sx={{ buttonNakedInheritStyle }}
```
AndreaCimini90, line 63: This style maybe is useless, add only verticalAlign: 'unset' and check that mobile and desktop views are ok

packages/pn-commons/src/components/EmptyState.tsx
```diff
@@ -47,29 +49,25 @@ function EmptyState({
...
-        </> 
-      }
+            <Typography color="primary" variant="body2" fontWeight={'bold'}>
+              {emptyActionLabel}
+            </Typography>
+          </ButtonNaked>
+        </>
+      )}
       &nbsp;
       <Typography variant="body2" sx={{ display: 'inline' }}>
```
AndreaCimini90, line 72: Add secondaryMessage.emptyMessage && component. Check that this doesn't cause problem around the application.

AndreaCimini90: The same for secondaryMessage.emptyActionLabel. Check that this doesn't cause problem around the application.

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #680, opened 2023-04-07
reworked paymentInfo flux when payment history is not empty - pn-5118

AndreaCimini90:
## Short description
When payment history is not empty, we must avoid the paymentInfo api call

- NotificationPayment pf and pg -> changed the condition to show the payment history box and manage the paymentInfo flux as if payment succeed

## How to test
Log with pf pointing to coll environment -> user Louis Amstrong -> notification MNKT-UXVA-QZDJ-202303-A-1 -> check that payment history box is shown and paymentInfo box has the success status


packages/pn-personafisica-webapp/src/__test__/test-utils.tsx
```diff
@@ -37,8 +39,51 @@ const axe = configureAxe({
...
+ * Utility function to mock api response
+ * @param client Axios client
+ * @param method the api method: 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH'
+ * @param path the api path
+ * @param code the response code
+ * @param request body request
+ * @param response response
+ * @returns the mock instance
+ */
+function mockApi(
```
ZorbaDimatteo, line 52: @AndreaCimini90 do you think is possibile to put this mock function in pn-commons?

ZorbaDimatteo (approved):

PR #683, opened 2023-04-12
Replaced text about notification viewed status

NicolaGiornetta:
## Short description
Replaced text about notification viewed status.

- Text from 'Perfezionata per visione' to 'Avvenuto accesso'

## How to test
Find any notification with status VIEWED. You should see 'Avvenuto accesso' now.


AndreaCimini90 (commented):
I'm not sure that the updates done are those requested.
In the be we have only the status VIEWED, but in the front-end we have two status:
- VIEWED for those notifications viewed before EFFECTIVE_DATE
- VIEWED_AFTER_DEADLINE for those notifications viewed after EFFECTIVE_DATE
what is the request of the issue?
There can be three choices:
 1. We must change the copy of the VIEWED state
 2. We must change the copy of the VIEWED_AFTER_DEADLINE state
 3. We must change the copy for both state
In the third case, does it still make sense to keep two distinct states?

AndreaCimini90 (approved):

PR #692, opened 2023-04-20
Pn 5260 fe come operatore della pa mittente voglio vedere il codice identificativo dei gruppi associati a una singola api key

NicolaGiornetta:
## Short description
Added ability to view Group IDs in API Key list.

- Added / removed strings from apikeys.json
- Added tooltip in API Key groups which shows their IDs
- Added option for subtitle position in API Key modal component
- Replaced CopyToClipboard common component with MUI-Italia's CopyToClipboardButton
- Added new menu item for context menu and new modal which shows all the Group IDs of API Key
- Added mapping function which merges infos from user groups to API Key list
- Added new parameter to CustomTagGroup which disables tooltip mouseover
- Fixed / added new tests

## How to test
Mouse over on groups and you should see tooltip with Group IDs. Also you should have ability to copy them.
Click context menu and select view groups id and modal opens with list of groups id. Also you should have ability to copy single group id.


packages/pn-pa-webapp/src/utils/apikeys.utility.tsx
```diff
@@ -2,7 +2,8 @@ import { Box } from '@mui/material';
 import { formatDate, isToday } from '@pagopa-pn/pn-commons';
```
AndreaCimini90, line 2: Write the test for the new function

packages/pn-pa-webapp/src/pages/components/ApiKeys/ApiKeyModal.tsx
```diff
@@ -4,7 +4,8 @@ import { Button, Grid, SxProps, Typography } from '@mui/material';
 type Props = {
```
AndreaCimini90, line 4: Write the test for the new component

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #702, opened 2023-04-26
feat(pn-5378): nuova validazione domini pec

LorenzoCocciarelli:
## Short description
This PR brings the changes required by the new validation of PEC emails, now if the user validates through code the validity of the email, but the system has not received the delivery confirmation of the email, the app shows a dialog window to inform the user that the process of PEC validation is in progress, when a user lands on the 'Recapiti' page, if the process is still happening, the text 'Verifica in corso' is shown, adjacently there's button to undo/stop the process, this button currently does something only app side, it shows a modal to inform the user that to stop the process, a new PEC email should be supplied and removes the not validated (notValidPec: true) from the redux state and the 'Verifica in progress' from the interface, however reloading the page will show the message again.

- Adds modal to show the start of the PEC validation progress
- Adds button and modal to remove the 'Verifica in corso' message
- Adds handling for the different 204 status responses
- Adds new error handling for code verification API call
- Adds new copy lines for error messages, modals and validation status
- Aligns tests with new behaviour

## How to test
- Go to Recapiti page, add a legal contact and insert the code the email should be verified within 20 seconds
- Go to Recapiti page, add a legal contact ending in @fail.it (or fail.anything) and insert the code the email should not be verified and will be deleted after 24 hours


ZorbaDimatteo (changes requested):
@LorenzoCocciarelli 
remove old un-referenced labels for error messages (eg. wrong-code in recapiti.json)
and check also the case where codeValid is false and value is null for legal contact

PR #707, opened 2023-05-02
Fixed non-refreshing status enabled disabled app IO in contacts

NicolaGiornetta:
## Short description
Fixed non-refreshing status enabled disabled app IO in contacts.

- Added useState vars instead of common const for status

## How to test
You should be able to see changing status enable / disable button in IO contac box now.


AndreaCimini90:
I'm not sure about the solution proposed, because the component should be re-rendered each time redux state changes; due to the re-rendering, the status variable should be ravaluated and the UI appereance should change

AndreaCimini90 (approved):

PR #713, opened 2023-05-04
Added tax code string handler between PG and PF

NicolaGiornetta:
## Short description
In the PA app, customize several labels regarding the kind of tax id, namely CodiceFiscale or Partita IVA.

- Notification detail: shows label "Codice Fiscale destinatario" or "Partita IVA destinatario" depending on the value of the tax id (so that if the tax id is a 16-digit CF, the "Codice Fiscale" label is shown even if the recipient is a persona giuridica).
- Notification creation, second step - recipient: the label/placeholder for tax id is "Codice Fiscale destinatario" if the recipient kind is persona fisica, "Partita IVA o Codice Fiscale destinatario" if it's persona giuridica. Also, the error in case of duplicated tax id in several recipients changes to "... Codice Fiscale / Partita IVA ...".
- In the dashboard, now the subtitle reads "... Puoi filtrarle per Codice Fiscale o Partita IVA, ...". 

## How to test
Always in PA:
- open a single-recipient notification detail for a PIva-identified PG, you should see label "Partita IVA" before tax id.
- open a single-recipient notification detail for a PF, you should see label "Codice fiscale destinatario" before fiscal code.
- open the dashboard and look at the subtitle.
- proceed to the notification creation, go to the second step, change the persona giuridica kind, type wrong or duplicated tax ids.


clombardi (approved):
OK to merge

PR #722, opened 2023-05-10
Fix/pn 5758 group admin should not see contacts

LorenzoCocciarelli:
## Short description
Adds check for recapiti item in sidebar and makes recapiti a private route for group admin users in pg-webapp

- Adds check for recapiti item in sidebar and hides it for group admin users
- Makes recapiti a private route for group admin users

## How to test
Login pg app with a group admin user, check that recapiti item in sidebar is not visible and route `recapiti` answers with a 'page not found'


packages/pn-personagiuridica-webapp/src/component/DomicileBanner/DomicileBanner.tsx
```diff
@@ -36,10 +37,10 @@ const DomicileBanner = () => {
   };
 
   useEffect(() => {
-    if (legalDomicile && legalDomicile.length > 0) {
+    if ((legalDomicile && legalDomicile.length > 0) || isGroupAdmin ) {
```
AndreaCimini90, line 40: I sohuld revome the check here and I sohuld put the condition {isGroupAdmin && <DomicileBanner...} where the component is used

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #726, opened 2023-05-10
PN-5484 - add copy to analog flow events with legal facts attached

clombardi:
## Short description
As an addenda to PN-5484, it was decided to show "C'è un nuovo documento allegato." as copy for the timeline events (mostly SEND_ANALOG_PROGRESS) having attached a legal fact.

- Just modified the notifiche.json accordingly.

## How to test
Enter an analog flow notification in PA / PF / PG, open detail for the status "Invio in corso". The events including an attachment just have the indicated copy.  
Example: ambiente DEV, IUN VZWU-ULXW-TRMG-202305-Y-1, Comune di Palermo (the recipient is Maria Montessori, but she hasn't accepted the TOS and it should be kept that way). For PF, most probably cristoforocolombo will have several analog flow notifications (you can see "analogico" in the description).


ZorbaDimatteo (changes requested):
@clombardi since the copy is the same for each case ("C'è un nuovo documento allegato.")
is it necessary to provide all this constants and labels?

clombardi:
> @clombardi since the copy is the same for each case ("C'è un nuovo documento allegato.") is it necessary to provide all this constants and labels?

@ZorbaDimatteo It could be done in a different way, but it would put a bit more logic in the code. The current solution relies on a single rule to compute the copy of any analog-flow-related timeline events, based on the `deliveryDetailCode` (RECRN001B, RECRN002B, ecc.), hence whatever the same copy must be used for different values of `deliveryDetailCode`, it must be copied for each of them.  
We could decide instead a single entry in the i18n file for all values of `deliveryDetailCode` for which no specific description is set, or else define two sets of allowed values of `deliveryDetailCode` instead of just one: one set with specific copies, a second set with a common copy.  
I guess that a 5-minute huddle should suffice to define a strategy.

ZorbaDimatteo (approved):

PR #727, opened 2023-05-10
fix: insert a11y in contacts component

leleOFA:
## Short description
insert a11y in contacts page 

- insert in every description of component `aria` attribute 

## How to test
active voicOver in our MacOS and slide all page with tab button, of keyboard.

### DIFF
--- a/packages/pn-commons/src/components/TitleBox.tsx
+++ b/packages/pn-commons/src/components/TitleBox.tsx
@@ -51,7 +51,6 @@ export default function TitleBox({
       {title && (
         <Grid id="item" item xs={12} mb={mbTitle}>
           <Typography
-            component={'h1'}
             id="title-of-page"
             role="heading"
             aria-label={ariaLabel}
--- a/packages/pn-commons/src/components/TitleBox.tsx
+++ b/packages/pn-commons/src/components/TitleBox.tsx
@@ -20,6 +20,8 @@ type Props = {
   sx?: SxProps<Theme>;
   /** Children */
   children?: ReactFragment;
+  /** a11y for component */
+  ariaLabel?: string;
 };
 
 /**
@@ -34,18 +36,25 @@ export default function TitleBox({
   variantTitle = 'h1',
   variantSubTitle = 'h5',
   sx,
+  ariaLabel,
   children,
 }: Props) {
   return (
-    <Grid id="container" aria-orientation="horizontal" tabIndex={0} container mt={mtGrid} sx={sx}>
+    <Grid
+      id="page-header-container"
+      aria-orientation="horizontal"
+      tabIndex={0}
+      container
+      mt={mtGrid}
+      sx={sx}
+    >
       {title && (
         <Grid id="item" item xs={12} mb={mbTitle}>
           <Typography
             component={'h1'}
-            key={'recapiti-title'}
-            id={'title'}
+            id="title-of-page"
             role="heading"
-            aria-label={title.toString()}
+            aria-label={ariaLabel}
             aria-selected="true"
             variant={variantTitle}
           >
--- a/packages/pn-commons/src/components/TitleBox.tsx
+++ b/packages/pn-commons/src/components/TitleBox.tsx
@@ -37,16 +37,34 @@ export default function TitleBox({
   children,
 }: Props) {
   return (
-    <Grid container mt={mtGrid} sx={sx}>
-      {title && <Grid item xs={12} mb={mbTitle}>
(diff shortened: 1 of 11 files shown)


packages/pn-commons/src/components/TitleBox.tsx
```diff
@@ -37,16 +37,34 @@ export default function TitleBox({
...
-      <Grid item xs={12}>
+    <Grid id="container" aria-orientation="horizontal" tabIndex={0} container mt={mtGrid} sx={sx}>
+      {title && (
+        <Grid id="item" item xs={12} mb={mbTitle}>
+          <Typography
+            component={'h1'}
+            key={'recapiti-title'}
+            id={'title'}
+            role="heading"
+            aria-label={title.toString()}
```
NicolaGiornetta, line 48: Not a good idea using `toString()` method. It gives "`[object Object]`" if you use react nodes instead of string in `title` parameter. is better to use a separate optional prop such ariaLabel: string instead.

packages/pn-commons/src/components/TitleBox.tsx
```diff
@@ -37,16 +37,34 @@ export default function TitleBox({
...
-      {title && <Grid item xs={12} mb={mbTitle}>
-        <Typography variant={variantTitle}>{title}</Typography>
-      </Grid>}
-      {subTitle && <Grid item xs={12} mb={mbSubTitle}>
-        <Typography variant={variantSubTitle} sx={{ fontSize: '18px' }} component={typeof subTitle !== "string" ? "div" : "p"}>
-          {subTitle}
-        </Typography>
-      </Grid>}
-      <Grid item xs={12}>
+    <Grid id="container" aria-orientation="horizontal" tabIndex={0} container mt={mtGrid} sx={sx}>
```
NicolaGiornetta, line 40: Not sure about this id prop. The value `container` is too generic but also what's purpose?

packages/pn-commons/src/components/TitleBox.tsx
```diff
@@ -37,16 +37,34 @@ export default function TitleBox({
...
+      {title && (
+        <Grid id="item" item xs={12} mb={mbTitle}>
+          <Typography
+            component={'h1'}
+            key={'recapiti-title'}
+            id={'title'}
```
NicolaGiornetta, line 46: Same as `container` id.

NicolaGiornetta (changes requested):
There are some change requests I made. You should also do the same work `for personagiuridica`.

packages/pn-commons/src/components/TitleBox.tsx
```diff
@@ -37,16 +37,34 @@ export default function TitleBox({
...
-          {subTitle}
-        </Typography>
-      </Grid>}
-      <Grid item xs={12}>
+    <Grid id="container" aria-orientation="horizontal" tabIndex={0} container mt={mtGrid} sx={sx}>
+      {title && (
+        <Grid id="item" item xs={12} mb={mbTitle}>
+          <Typography
+            component={'h1'}
+            key={'recapiti-title'}
```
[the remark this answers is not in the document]
leleOFA: delete key attributes

packages/pn-commons/src/components/TitleBox.tsx
leleOFA: correct with different id

packages/pn-commons/src/components/TitleBox.tsx
leleOFA: correct as before

packages/pn-personafisica-webapp/src/pages/Contacts.page.tsx
```diff
@@ -97,6 +97,7 @@ const Contacts = () => {
             title={t('title')}
             subTitle={subtitle}
             variantSubTitle={'body1'}
+            ariaLabel="Recapiti"
```
NicolaGiornetta, line 100: Use the translation hook, same as `title`.

NicolaGiornetta (changes requested):
You have to add ariaLabel prop in TitleBox in PG contacts page.

Check for CourtesyContacts, InsertLegalContact and LegalContactList in PF, you missed the a11y in these files as you did in PG.

Some tests are failing, please check it up.

packages/pn-personafisica-webapp/src/pages/Contacts.page.tsx
leleOFA: done

NicolaGiornetta (approved):
Ok to merge.

PR #743, opened 2023-05-16
manage error with response null - pn-5888

AndreaCimini90:
## Short description
Errors that has response null weren't managed and so the interface didn't react as expected. Now these errors are managed as generic (500) errors.

- updated redux.utility.ts
- localized empty states in PF and PG

## How to test
- log with PG, PF or PA -> go to notification list -> change the path of the notification list api -> check that toast and empty state are shown


clombardi (approved):
Verified and tested, ready to merge.

PR #746, opened 2023-05-16
fix: remove border from sidebar items when they are not selected

LorenzoCocciarelli:
## Short description
The border in the sidemenu is visible only for the selected item, for other items has width 0px.

- Changed the border for sidemenu items to not be visible

## How to test
Check that in high contrast (prefereably windows) the border in the sidemenu is visible only for the selected item for any webapp


LorenzoCocciarelli:
moved fix to mui-italia

PR #747, opened 2023-05-16
PN-5875 - notification subject controlled length in manual notification creation

clombardi:
**Beware**  
The issue should pass direct to "In Test"when both this PR and the parallel work on the backend have been merged. Please inform @clombardi when merging this PR, instead of passing the issue to "In review". Thanks a lot!!

## Short description
Changed the validation of the subject field in the first step (preliminary information) in manual notification creation, to enforce the length to be no less than 10 and no more than 120.  

Just changed the specification of the validation on `subject`, and added a test.

## How to test
Just enter any PA user, enter the manual notification creation procedure, and play with the length of the "Oggetto della notifica" field in the first step.

### DIFF
--- a/packages/pn-pa-webapp/src/pages/components/NewNotification/PreliminaryInformations.tsx
+++ b/packages/pn-pa-webapp/src/pages/components/NewNotification/PreliminaryInformations.tsx
@@ -60,7 +60,7 @@ const PreliminaryInformations = ({ notification, onConfirm }: Props) => {
 
   const validationSchema = yup.object({
     paProtocolNumber: requiredStringFieldValidation(tc, 256),
-    subject: requiredStringFieldValidation(tc, 134),
+    subject: requiredStringFieldValidation(tc, 120).min(10, tc('too-short-field-error', { minLength: 10 })),
     abstract: yup.string().max(1024, tc('too-long-field-error', { maxLength: 1024 })).matches(dataRegex.noSpaceAtEdges, tc('no-spaces-at-edges')),
     physicalCommunicationType: yup.string().required(),
     paymentMode: yup.string().required(),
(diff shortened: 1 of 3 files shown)


packages/pn-pa-webapp/src/pages/components/NewNotification/PreliminaryInformations.tsx
```diff
@@ -60,7 +60,7 @@ const PreliminaryInformations = ({ notification, onConfirm }: Props) => {
 
   const validationSchema = yup.object({
     paProtocolNumber: requiredStringFieldValidation(tc, 256),
-    subject: requiredStringFieldValidation(tc, 134),
+    subject: requiredStringFieldValidation(tc, 134).min(10, tc('too-short-field-error', { minLength: 10 })),
```
AndreaCimini90, line 63: Can you put the min check into requiredStringFieldValidation function as done for max?

AndreaCimini90 (changes requested):

packages/pn-pa-webapp/src/pages/components/NewNotification/PreliminaryInformations.tsx
clombardi: Ciao @AndreaCimini90 I preferred to avoid that since `subject` is the *only* field including a `min` check, while `requiredStringFieldValidation` is used ~10 times.

clombardi: Even when I don't agree, I will proceed as suggested given @AndreaCimini90 still thinks it's the best thing to do.

AndreaCimini90 (approved):

PR #757, opened 2023-05-18
Pn 5872 fe distinguere gli aar in caso di destinatari multipli mittente

NicolaGiornetta:
## Short description
Added recipients data in PA next to AAR documents name.

- Added recipient denomination and tax id next to AAR document name

## How to test
Open a multi recipient notification in PA and look for Other documents box. You should see AAR title documents with recipient name and tax id.


AndreaCimini90 (commented):
Add a test for the new use case

AndreaCimini90 (approved):

PR #762, opened 2023-05-19
fix: add group-admin empty state message for notifications in pg

LorenzoCocciarelli:
## Short description
Change the message for the empty state in Notifications page in case the user is a group admin one, removes link to contacts page

- Adds function to get first message
- Adds new entry for group admin empty state message in notifiche.json

## How to test
Check that for a user with empty state the message doesn't show the link to recapiti page


AndreaCimini90 (changes requested):

LorenzoCocciarelli:
solved in PN-5877

PR #763, opened 2023-05-19
Prevent notiifcation when 451 error occur - pn-5812

AndreaCimini90:
## Short description
When the user is not allowed to log in, the error must not be notified through toast

- Added forbidden error handler in SessionGuard.tsx (PF/PG/PA)

## How to test
PF - login with Dino and check that no toast is shown
PA, PG - do the same test


clombardi (approved):
OK to merge, actually tested in PF / PA / PG

PR #766, opened 2023-05-22
PG - filter delegation by taxId - pn-5795

AndreaCimini90:
## Short description
In PG, delegation are now filterable by taxId

- Deprecated api that returns all the names of the delegators
- Replaced filter by name with filter by taxId

## How to test
PG -> login with DanteAlighieri - DivinaCommedia -> go to "Deleghe a carico" -> filter by taxId and check that delegations are filtered correctly


NicolaGiornetta (changes requested):
When you do filters with any parameters but taxId, the API responds with VALIDATION_ERROR. This because parameter taxId is sent as empty string. A possible solution is to conditionally send taxId parameter only if is filled.

PR #770, opened 2023-05-23
fix: insert box shadow in autocomplete

leleOFA:
## Short description
in the application, the mui component used Autocomplete don't had box shadow when explode options to selection

- insert in autcomplete component Paper component for render options with box-shadow same as every mui component

## How to test
when explode options to selection in autocomplete you will view box shadow


AndreaCimini90 (changes requested):
We repeat the same code in all parts where we use the Autocomplete.
I think that this change should be done at theme level i.e. in mui-italia

AndreaCimini90 (approved):

PR #783, opened 2023-05-29
fix: improve coverage for delegation.utility

LorenzoCocciarelli:
## Short description
Improves coverage in pf and pg for delegation.utility file

## How to test
Run tests, they should not break


packages/pn-personagiuridica-webapp/src/redux/delegation/types.ts
```diff
@@ -0,0 +1,79 @@
+import { RecipientType } from '@pagopa-pn/pn-commons';
```
AndreaCimini90, line 1: Pay attention. This models are already defined. They were grouped in another file during the development of the delegation part

AndreaCimini90 (changes requested):

clombardi (approved):
OK, ready to merge

PR #790, opened 2023-06-05
fix: sizes of new api key form

alessandrogelmi:
## Short description
Fix size of title in new api key form and fix label's margins.

- Fix size of form title
- Fix label's margins
- Mapped form title in locales

## How to test
Go to new api key and will see the new form title size and label's margins

### DIFF
--- a/packages/pn-pa-webapp/src/pages/NewApiKey.page.tsx
+++ b/packages/pn-pa-webapp/src/pages/NewApiKey.page.tsx
@@ -128,7 +128,9 @@ const NewApiKey = () => {
                   </Typography>
                   <Box>
                     <Paper sx={{ padding: '24px', marginTop: '40px' }} className="paperContainer">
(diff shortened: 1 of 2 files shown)


packages/pn-pa-webapp/src/pages/NewApiKey.page.tsx
```diff
@@ -128,7 +128,9 @@ const NewApiKey = () => {
                   </Typography>
                   <Box>
                     <Paper sx={{ padding: '24px', marginTop: '40px' }} className="paperContainer">
-                      <Typography variant="h5">Altre informazioni</Typography>
+                      <Typography component="h3" variant="h6">
```
clombardi, line 131: Wouldn't be better to use the `SectionHeading` component from `pn-commons`?

clombardi (changes requested):
Hi Alessandro, just a little detail to avoid the duplication of style specs.

PR #796, opened 2023-06-07
fix: change name in revocation modal from user name to organization name

LorenzoCocciarelli:
## Short description
Change name in revocation modal from user name to organization name

## How to test
Check that in the delegation revocate modal the organization name is shown instead of the user name


clombardi (approved):
OK, ready to merge

PR #827, opened 2023-06-14
fix: deleted groups handling

alessandrogelmi:
## Short description
Deleted groups handling. Show only active groups when insert/modify a new notification, api key and delegations.

- Deleted groups handling

## How to test
On PA:
- When yuo create a new notification or a new api key you should see only active groups
On PG:
- When accept a new delegation you can assign only to an active group
- When you filter delegation of the company you should see only active groups

### DIFF
--- a/packages/pn-personagiuridica-webapp/src/component/Deleghe/AcceptDelegationModal.tsx
+++ b/packages/pn-personagiuridica-webapp/src/component/Deleghe/AcceptDelegationModal.tsx
@@ -25,6 +25,7 @@ import {
 import { ServerResponseErrorCode } from '../../utils/AppError/types';
 import { useAppSelector } from '../../redux/hooks';
 import { RootState } from '../../redux/store';
+import { GroupStatus } from '../../models/groups';
 
 type Props = {
   isEditMode: boolean;
@@ -180,7 +181,7 @@ const AcceptDelegationModal: React.FC<Props> = ({
             id="groups"
             size="small"
             fullWidth
-            options={groups}
+            options={groups.filter((group) => group.status !== GroupStatus.DELETED)}
             disableCloseOnSelect
             multiple
             noOptionsText={t('deleghe.table.no-group-found')}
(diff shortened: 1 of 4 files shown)


packages/pn-personagiuridica-webapp/src/models/groups.ts
```diff
@@ -1,6 +1,7 @@
 export enum GroupStatus {
   ACTIVE = 'ACTIVE',
   SUSPENDED = 'SUSPENDED',
+  DELETED = 'DELETED',
```
clombardi, line 4: I guess the DELETED status should be added to `pn-pa-webapp` as well.

packages/pn-personagiuridica-webapp/src/component/Deleghe/DelegationsOfTheCompany.tsx
```diff
@@ -392,7 +393,7 @@ const DelegationsOfTheCompany = () => {
                   id="groups"
                   size="small"
                   fullWidth
-                  options={groups}
+                  options={groups.filter((group) => group.status !== GroupStatus.DELETED)}
```
clombardi, line 396: Did you confirm how we should handle SUSPENDED status (if it still exists)?

packages/pn-personagiuridica-webapp/src/component/Deleghe/AcceptDelegationModal.tsx
```diff
@@ -180,7 +181,7 @@ const AcceptDelegationModal: React.FC<Props> = ({
             id="groups"
             size="small"
             fullWidth
-            options={groups}
+            options={groups.filter((group) => group.status !== GroupStatus.DELETED)}
```
clombardi, line 184: Did you confirm how we should handle SUSPENDED status (if it still exists)?

clombardi (changes requested):
Even when the issue indicates to filter only the DELETED status, I still have doubts whether the SUSPENDED status should be filtered out as well.  
The SUSPENDED status is still present backend-side, cfr. https://github.com/pagopa/pn-external-registries/blob/develop/docs/openapi/pn-selfcare-external-v1.yaml. 

So please make yourself sure about this aspect (and generate a coherent definition of `GroupStatus` for both PA and PG) before merging this PR. Thanks!

clombardi (approved):
now OK, ready to commit

PR #836, opened 2023-06-16
Fix/pn 6087

leleOFA:
## Short description
fix style of card in pa

- bold pre-label
- align in left chip of status
- insert tag style in gruop

## How to test
see card same as figma


ZorbaDimatteo (changes requested):
@leleOFA apply same fix to notifications cards for pg and pf

ZorbaDimatteo (approved):

PR #867, opened 2023-06-28
fix: a11y of landing page

alessandrogelmi:
## Short description
Fix accessibility of landing page

- Added tabIndex and aria label to every text componenti (excepts for titles)

## How to test
Go to the landing page and try to navigate it with TAB and voiceover


alessandrogelmi:
@ZorbaDimatteo titles of every MUI Italia component doesn't support JSX element (accept only string). So, I can't pass the tabIndex and aria-label props.

PR #877, opened 2023-07-03
PN-6828 - allowed particular accessibility link for one URL pattern / disabled languages other than Italian

clombardi:
## Short description
Allowed a different destination URL for the "Accessibilità" link inside the footer, depending on the particular URL used to access the landing site.  
Currently the accessibilità URL changes to  
https://form.agid.gov.it/view/ee21027b-0e20-4e57-98c1-b3652daa6ef5/  
if the landing site URL includes the pattern  
`notifichedigitali.it`  
(like in https://www.notifichedigitali.it). But this can be changed via env variables.

On the other hand, all the languages other than Italian were taken out from the language selection, also present in the banner. The language selection is still active, but allows to choose only Italian.

- Changed the definition of `preLoginLinks`, now is a function which takes the windowURL as parameter. This is passed in the (now) call inside `LandingLayout`.
- Added env variables `NEXT_PUBLIC_ACCESSIBILITY_PARTICULAR_LANDING_URL_PATTERN` and `NEXT_PUBLIC_ACCESSIBILITY_PARTICULAR_ACCESSIBILITY_URL`.
- For the accessibility link, if the window URL includes the `PARTICULAR_LANDING_URL_PATTERN`, then the `PARTICULAR_ACCESSIBILITY_URL` is used instead of the default value https://form.agid.gov.it/view/eca3487c-f3cb-40be-a590-212eafc70058/.
- Commented out all the languages other than Italian in `LANGUAGES`.

## How to test
Enter the landing site with both https://www.notifichedigitali.it and other URL. You should arrive to different URLs when you click on "Accessibilità" in the fourth column of the footer. For https://www.notifichedigitali.it the obtained page should refer to 30.06.2023, for other URLs to 23.09.2022 instead.  
Of course you could change https://www.notifichedigitali.it with a different URL, just changing the value of `NEXT_PUBLIC_ACCESSIBILITY_PARTICULAR_LANDING_URL_PATTERN` in the file `.env`.

Click on the language selection bar, also in the fourth column of the footer. You should find "Italiano" as the only available option.


packages/pn-landing-webapp/src/layout/LandingLayout.tsx
```diff
@@ -22,6 +22,20 @@ const LandingLayout = ({ children }: Props) => {
   const { pathname } = useRouter();
   const appData = getAppData();
   const assistanceEmail = pathname !== "/pubbliche-amministrazioni" ? SEND_PF_HELP_EMAIL : PAGOPA_HELP_EMAIL;
+  const [windowURL, setWindowURL] = useState<string>();
+
+  // 
+  useEffect(() => {
+    if (typeof window !== 'undefined') {
+      setWindowURL(window.location.origin);
+      // console.log('asserting window.location.origin');
```
AndreaCimini90, line 31: remove commented code

packages/pn-landing-webapp/src/layout/LandingLayout.tsx
```diff
@@ -22,6 +22,20 @@ const LandingLayout = ({ children }: Props) => {
   const { pathname } = useRouter();
   const appData = getAppData();
   const assistanceEmail = pathname !== "/pubbliche-amministrazioni" ? SEND_PF_HELP_EMAIL : PAGOPA_HELP_EMAIL;
+  const [windowURL, setWindowURL] = useState<string>();
+
+  // 
+  useEffect(() => {
```
AndreaCimini90, line 28: are the useEffect and useState necessary? I think that you can define a const windowURL = window ? window.location.origin : null

AndreaCimini90 (changes requested):

packages/pn-landing-webapp/src/layout/LandingLayout.tsx
clombardi: I was inspired by the second response in https://stackoverflow.com/questions/55151041/window-is-not-defined-in-next-js-react-app, as suggested by @ZorbaDimatteo . I fear to access window in arbitrary places, since I don't fully understand which part of the code is executed server-side and which part browser-side.

packages/pn-landing-webapp/src/layout/LandingLayout.tsx
clombardi: done

AndreaCimini90 (approved):

PR #883, opened 2023-07-04
Pn 6315 ds bug ux mittente tooltip campo codice tassonomico

NicolaGiornetta:
## Short description
Added icon info with tooltip to taxonomy code input in preliminary information.

- Added icon with tooltip
- Added new copy to notifiche.json in PA

## How to test
Login PA then create manual notification. Hover mouse to icon info in taxonomy code input. You should see a tooltip now.


packages/pn-pa-webapp/src/pages/components/NewNotification/PreliminaryInformations.tsx
```diff
@@ -117,6 +119,12 @@ const PreliminaryInformations = ({ notification, onConfirm }: Props) => {
     fetchGroups();
   }, [fetchGroups]);
 
+  const InfoTooltip = ({ tooltip }: { tooltip: string | ReactNode }) => (
```
clombardi, line 122: I suggest to  embed the `Tooltip` inside a `InputAdornment position='end'` element. Besides being more aligned with MUI documentation, in this way the `InfoOutlined` icon appears with color #5C6F82 as indicated in the Figma document
image
Note that if `InputAdornment` is not used, then the color changes to #17324D, the one used for texts.

clombardi (changes requested):
The solution is OK in general terms. I suggest two modifications, one regarding the i18n files for the different languages as discussed earlier today in the #team-frontend-pn Slack channel, the other regarding a detail about the adornment implementation.

clombardi (approved):
OK, ready to merge.

PR #899, opened 2023-07-07
Fix always shown special contacts box in contacts page

NicolaGiornetta:
## Short description
Fix always shown special contacts box section in contacts page PF.

- Added a check about App IO

## How to test
This bug only affects in PROD because of App IO and with this fix should be gone.


packages/pn-personafisica-webapp/src/pages/Contacts.page.tsx
```diff
@@ -102,6 +102,15 @@ const Contacts = () => {
     </>
   );
 
+  const courtesyContactsNotEmpty = () => {
```
clombardi, line 105: Hi Nicola, I guess that the condition is not written correctly. Assume that the user has just one courtesy address like the following
```
{ channelType: "APPIO", value: "DISABLED" }
```
The first condition wouldn't be met (since `courtesy[0].value === 'ENABLED')` wouldn't hold) but the second condition reads simply `courtesy.length > 0`, and that's obviously true. Therefore `courtesyContactsNotEmpty()` would evaluate to `true`, and hence the special contacts list would be shown (I write down all the sequence so you can confirm or question it).  

I suggest to change the implementation of `courtesyContactsNotEmpty`, by defining a `const onlyIrrelevantAddresses` that checks that all addresses are irrelevant, and return `!onlyIrrelevantAddresses`.  
We define an address to be irrelevant if `address.channelType === CourtesyChannelType.IOMSG && address.value === 'DISABLED'`. You can use `Array.every` to check that all addresses are irrelevant (and thus considering the eventual case in which more than one useless address shows up for a single user).

Please add jest tests about this detail, since we cannot actually test it on DEV / TEST / UAT. I would write at least four tests
- no addresses => no special contacts.
- only one disabled IO address => no special contacts.
- only one enabled IO address => special contacts.
- a disabled IO address + a mail address => special contacts.

clombardi (changes requested):
I'm afraid the implementation proposed is not correct, I explain in detail in the specific comment. So if I'm right, the PR should not be merged in its current form.

packages/pn-personafisica-webapp/src/pages/Contacts.page.tsx
clombardi: OK, now it works like a charm

clombardi (approved):
OK, ready to merge

PR #901, opened 2023-07-07
fix(PN-7095): domicile banner consider types actually lacking

clombardi:
## Short description
Changed how the `DomicileBanner` computes the candidate channel types (among IO / SMS / PEC / mail for PF, SMS / PEC / mail for PG) from which select the channel type to be shown.  
After this PR, the candidate channel types are those for which the logger used has not set a default address. Even if the PEC is set, the courtesy channels for which no address is set are indeed candidates, this differ from the previous implementation in which the presence of a PEC blocked the banner to appear no matter which courtesy addresses were or weren't set.  

The functional details can be checked in the JIRA issue https://pagopa.atlassian.net/browse/PN-7095.

**Note 1.1**:  
The addresses eventually added and/or removed in the current browser session _are not considered_, in order to update the information considered by the banner you must reload.

**Note 1.2**:  
The PEC is considered as set even if `pecValid` is false, because this means that the PEC is under analysis, so that the user has registered it.

- Changed the addresses kept in the API call that retrieves the logged user' addresses launched upon PF / PG app start. Now it keeps all default (i.e. not aimed to specific senders) addresses, not only legal ones.
- Changed the name of the property that holds this information in the Redux state, from `legalDomicile` to `defaultAddresses`.
- Changed the logic of `DomicileBanner` to compute the lacking channel types, and select randomly one of them to show the banner. Also the condition to not show the banner is changed, now is the absence of any lacking channel types (i.e. `lackingChannelTypes.length === 0`).
- Changed the keys of the i18n texts for the banner, now they refer to specific channel types (e.g. `domicile_PEC`) instead of just to an arbitrary index (e.g. `domicile_1`).
- Performed the needed adaptations on tests.

## How to test
- Enter PF and/or PG with users having different (but not all) default addresses set, access *several times* (I tried with 20) to the list of notifications, verify that the upper domicile banner is present, and that the channel types proposed coincide with the lacking ones.
- Enter PF and/or PG with users having all default addresses set, the list of notification shouldn't show the upper domicile banner.

**Note 3.1**  
I guess that the easiest way to test different scenarios is to mock `Contacts.api` like this
```
  getDigitalAddresses: (): Promise<DigitalAddresses> => Promise.resolve({
    legal: [{
      addressType: 'legal',
      channelType: LegalChannelType.PEC,
      recipientId: 'id-1',
      senderId: 'default',
      value: 'popotitos@pec.it',
      codeValid: true,
      pecValid: true,
    }],
    courtesy: [{
      addressType: 'courtesy',
      channelType: CourtesyChannelType.EMAIL,
      recipientId: 'id-1',
      senderId: 'default',
      value: 'someOtherThing@normal.mail.it',
      codeValid: true,
    }],
  }),
```

**Note 3.2**  
At the moment the banner is shown for both own and delegated notifications, for both PF and PG. There is a separate bug to remove it from the delegated notification list if so decided, no need to report this.


AndreaCimini90 (approved):

packages/pn-personafisica-webapp/src/redux/sidemenu/actions.ts
```diff
@@ -21,7 +21,15 @@ export const getSidemenuInformation = createAsyncThunk<Array<Delegator>>(
...
+// PN-7095 - per capire quale categorie di recapito far vedere nel DomicileBanner
+// si devono prendere gli indirizzi default, tranne per AppIO che si prendono tutti (senza verificare default) 
+// il cui valore non sia DISABLED (cfr. description e commenti della issue JIRA)
 export const getDomicileInfo = createAsyncThunk<Array<DigitalAddress>>(
   'getDomicileInfo',
-  performThunkAction(async () => (await ContactsApi.getDigitalAddresses()).legal)
-);
+  performThunkAction(async () => {
+    const isDefaultAddress = (address: DigitalAddress) => 
+      address.senderId === 'default' || (address.channelType === CourtesyChannelType.IOMSG && address.value !== IOAllowedValues.DISABLED);
```
AndreaCimini90, line 31: Whit this condition also senderId = 'default' && channelType = IOMSG && value = DISABLED passes

AndreaCimini90 (changes requested):

packages/pn-personafisica-webapp/src/redux/sidemenu/actions.ts
clombardi: omg you're right

AndreaCimini90 (approved):

PR #908, opened 2023-07-11
Fixed issue with DomicileBanner in notification detail

NicolaGiornetta:
## Short description
This PR aims to fix a bug where DomicileBanner is shown in notification details as delegate.

- Added condition to show DomicileBanner both PF and PG

## How to test
DomicileBanner should not be shown anymore in notification detail as delegate.


clombardi (changes requested):
The changes proposed are OK, but unfortunately they're not enough. The domicile banner shouldn't appear on the notification list page as well, check e.g. the image below, in that page the banner shouldn't appear. This for both PF and PG.

Implementation detail: for PG you won't have a `mandateId` in the notification list page, but instead I guess that the `isDelegatedPage` prop will be useful for this PR.

image

clombardi (approved):
OK, ready to merge

PR #914, opened 2023-07-14
Fixed the filter that would have close even though the iun or recipie…

ImJimMoriarty:
…ntId was wrong

## Short description
The filter dialog on mobile was closing on submit even though one field such as iun and recipientId was not valid.

- Added a boolean prop for FilterNotificationsFormActions on each service to manage the dialog's closing system.
- Added a const boolean value that evaluates if the regex of the fields are correct.

## How to test
On mobile you can now type a wrong iun or recipientId and verify if it closes.


ImJimMoriarty:
Sorry, wrong issue it is actually this one: https://pagopa.atlassian.net/browse/PN-7120

ImJimMoriarty:
Changes: 

- Removed `isValid ` prop of FilterNotificationsFormActions from each module.
- Added `forwardRef` to CustomMobileDialogContent so that is able to forward the toggleOpen function and pass it to the parent component (FilterNotifications).
- Added `useRef` to FilterNotifications by importing the toggleOpen function.

PR #965, opened 2023-08-29
Added a statement into 'useDownloadDocuments' and removed new tab onc…

ImJimMoriarty:
## Short description
Added a new statement to the function `downloadDocument()` where it adds a `target blank` if you're in mobile view.

- Added a new statement into `downloadDocument()`

## How to test
You can check any NotificationDetail page inside PF/PG/PA and test with both different screen size (mobile and desktop).


packages/pn-commons/src/hooks/useDownloadDocument.ts
```diff
@@ -23,10 +24,10 @@ type Props = {
 // -------------------------------
 // Carlos Lombardi and Andrea Cimini, 2022.11.17
 // -------------------------------
-export function useDownloadDocument({ url, clearDownloadAction }: Props) {
+export function useDownloadDocument({ url, isMobile, clearDownloadAction }: Props) {
   useEffect(() => {
```
AndreaCimini90, line 28: Try to use the hook isMobile, to remove the isMobile property

AndreaCimini90 (changes requested):

packages/pn-commons/src/hooks/useDownloadDocument.ts
ImJimMoriarty: I've already tried the solution proposed and the result is an application crash. This is mainly caused by using a hook inside a regular function instead of a functional component.

ZorbaDimatteo:
@ImJimMoriarty vanno risolti i conflitti

ZorbaDimatteo (approved):

PR #980, opened 2023-09-13
Refactored login tests - pn 7406

AndreaCimini90:
## Short description
Refactored all the test of the login page

## How to test
Run login tests and check that everything goes well


packages/pn-personafisica-login/sonar-project.properties
```diff
@@ -1,6 +1,6 @@
 sonar.sources=src
-sonar.exclusions=src/**/__tests__/**, src/**/__test__/**
+sonar.exclusions=src/**/__tests__/**, src/**/__test__/**, src/__mocks__/**, src/utils/onetrust.ts
 sonar.test.inclusions=src/**/__tests__/**, src/**/__test__/**
 sonar.scm.exclusions.disabled=true
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
-sonar.coverage.exclusions=src/models/**, src/locales/**, src/assets/**, src/utils/**, src/**/__tests__/**, src/**/__test__/**
\ No newline at end of file
+sonar.coverage.exclusions=src/models/**, src/locales/**, src/assets/**, src/utils/**, src/**/__tests__/**, src/**/__test__/**, src/__mocks__/**, src/i18n.ts, src/index.tsx, src/utils/onetrust.ts, src/reportWebVitals.ts, src/setupTests.ts
```
ZorbaDimatteo, line 6: @AndreaCimini90 You added src/utils/onetrust.ts, but the exclusion rule src/utils/** was already present, which one should we keep?

packages/pn-personafisica-login/src/pages/login/Login.tsx
```diff
@@ -78,7 +79,7 @@ const Login = () => {
         email: undefined,
       }}
     >
-      <Grid container direction="column" my={isMobile ? 4 : 16}>
+      <Grid container direction="column" my={isMobile ? 4 : 16} data-testid="loginPage">
```
ZorbaDimatteo, line 82: @AndreaCimini90 if it's ok, we can use this test-id as id.
It would cover two issues: using ids for tests and put unique identifiers on page element

packages/pn-personafisica-login/sonar-project.properties
AndreaCimini90: ops i didn't see it

packages/pn-personafisica-login/src/setupTests.ts
```diff
@@ -5,20 +5,20 @@
...
-    "SPID_VALIDATOR_ENV_ENABLED": true,
-    "SPID_CIE_ENTITY_ID": "xx_servizicie_test",
-    "URL_API_LOGIN": "https://hub-login.spid.svil.pn.pagopa.it"
+  Configuration.setForTest<any>({
+    MIXPANEL_TOKEN: 'DUMMY',
+    ONE_TRUST_DRAFT_MODE: false,
+    ONE_TRUST_PP: 'mocked-id',
+    OT_DOMAIN_ID: 'fd5aef6f-a6d3-422b-87b7-aa5e2cb6510c',
+    PAGOPA_HELP_EMAIL: 'assistenza@pn.it',
+    PF_URL: 'https://portale.dev.pn.pagopa.it',
```
ZorbaDimatteo, line 16: I know that this file is used only for tests, but to avoid confusion I would use correct urls used as configuration to setupTests

ZorbaDimatteo (changes requested):
I left a comment to a page to refactor datatest-ids to ids. 
Let's consider to apply this idea to all newly added datatest-ids

packages/pn-personafisica-login/sonar-project.properties
AndreaCimini90: Removed the rule src/utils/** beacuse some files are tested

ZorbaDimatteo (approved):

PR #985, opened 2023-09-15
Removed unused method in AppError class

NicolaGiornetta:
## Short description
Cleaning AppError class.

- Removed method getErrorDetail() from AppError class. It was never used.


clombardi (approved):
The changes are OK. I do not perform the merge since I'm not sure about the target branch for each PR in the current situation of the project.

PR #988, opened 2023-09-21
Replaced RegEx of email

NicolaGiornetta:
## Short description
Replaced RegEx of email according the issue.

- string.utility.ts


ZorbaDimatteo (changes requested):
@NicolaGiornetta check the eslinter rule no-control-regex https://eslint.org/docs/latest/rules/no-control-regex and try to fix regex in a fashion that works for js and for web. If ASCII characteres are still necessaire, then we can leave them

ZorbaDimatteo:
@NicolaGiornetta some tests in PF/PG are failing, please fix them :)

ZorbaDimatteo (approved):

PR #989, opened 2023-09-22
Replaced regex for testPathIgnorePattern parameter in PA

NicolaGiornetta:
## Short description
This PR aims to fix a weird issue that causes all the tests in PA to fails in certain computers and OSs and it happens only in PA. In this case is on Windows 11 x64. The replacemente is simple. It's just another way to check for files with extensions .ts or .tsx

- package.json for PA

## How to test
Check if builds and tests are ok.

### DIFF
--- a/packages/pn-pa-webapp/package.json
+++ b/packages/pn-pa-webapp/package.json
@@ -41,8 +41,8 @@
   "scripts": {
     "start": "craco start",
     "build": "craco build",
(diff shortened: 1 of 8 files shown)


clombardi (approved):
Changes look harmless, and in effect kind of "solve" the problem shown noticed on Nicola' development envionment. Ready to test.

PR #993, opened 2023-09-25
Pn 7637 fe mittente filtro per codice iun

NicolaGiornetta:
## Short description
Added paste event control in filters for IUN and Fiscal Code / Tax ID.

- Added paste event for PA, PG and PF

## How to test
Open a detailed notification, then try to copy a IUN code by point the mouse a bit outside the row where IUN is located to get the whitespaces too. Then paste in filter, now should be fine.


clombardi (approved):
The change behaves as expected, found no flaws in the code, ready to merge

PR #1006, opened 2023-10-02
PN-8056 - used Trans component to show bold text in Api Keys's modals

alessandrogelmi:
## Short description
used Trans component to show bold text in Api Keys's modals

- Modified modals content on Api Keys page

## How to test
On PA go to Api Keys page and try to Rotate, block, delete, or activate an api key. You should see the api key name with bold weight

### DIFF
--- a/packages/pn-pa-webapp/src/pages/ApiKeys.page.tsx
+++ b/packages/pn-pa-webapp/src/pages/ApiKeys.page.tsx
@@ -264,9 +264,11 @@ const ApiKeys = () => {
               <ApiKeyModal
                 title={t('block-api-key')}
                 subTitle={
-                  <Trans i18nKey="block-warning1" values={{ apiKeyName: modal.apiKey?.name }}>
-                    {t('block-warning1', { apiKeyName: modal.apiKey?.name })}
-                  </Trans>
+                  <Trans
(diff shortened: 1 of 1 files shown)


packages/pn-pa-webapp/src/pages/ApiKeys.page.tsx
```diff
@@ -278,7 +278,11 @@ const ApiKeys = () => {
             {modal.view === ModalApiKeyView.ENABLE && (
               <ApiKeyModal
                 title={t('enable-api-key')}
-                subTitle={t('enable-warning', { apiKeyName: modal.apiKey?.name })}
+                subTitle={
+                  <Trans i18nKey="enable-warning" values={{ apiKeyName: modal.apiKey?.name }}>
+                    {t('enable-warning', { apiKeyName: modal.apiKey?.name })}
```
AndreaCimini90, line 283: The children in Trans component are used as default when translation fail (because key is not found).
This makes it difficult to understand which localizations are missing.
Do for all Trans used here.
Delete the default and try to use a wrong translation key, to confirm the above.

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1018, opened 2023-10-12
Fix copy paste duplication

NicolaGiornetta:
## Short description
Fixes copy paste duplication in filter form body.

- Fixed in PA, PF and PG

## How to test
Try to copy paste a TaxId in PA. Now it will be okay without duplication copy paste.


AndreaCimini90 (commented):
The PR is ok but i'ld write a test case (for each webapp) in which we try to copy/past a value with spaces

AndreaCimini90 (approved):

PR #1069, opened 2023-11-23
fix: moved documents above payments box and changed title variant

alessandrogelmi:
## Short description
Moved documents box above payments box and changed title variant of attached documents box 

- Moved documents box
- Changed title variant

## How to test
Go to a notification and you should see the attached documents box before payments box.
And the title of the attached documents box should be like the title of payments box

### DIFF
--- a/packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
+++ b/packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
@@ -392,6 +392,20 @@ const NotificationDetail = () => {
                   </Alert>
                 )}
                 <NotificationDetailTable rows={detailTableRows} />
+
+                <Paper sx={{ p: 3 }} elevation={0}>
+                  <NotificationDetailDocuments
+                    title={t('detail.acts', { ns: 'notifiche' })}
+                    documents={notification.documents}
+                    clickHandler={documentDowloadHandler}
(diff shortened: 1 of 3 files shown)


packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -415,17 +429,6 @@ const NotificationDetail = () => {
                 )}
 
                 {!mandateId && <DomicileBanner />}
```
AndreaCimini90, line 431: Check if also Domicile banner must be moved up

packages/pn-personagiuridica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -417,17 +431,6 @@ const NotificationDetail = () => {
                 )}
 
                 {visibleDomicileBanner() && <DomicileBanner />}
```
AndreaCimini90, line 433: Same as pf

AndreaCimini90 (changes requested):
A similar layout is applied to the pa version of the page. Why no update was needed?

alessandrogelmi:
> A similar layout is applied to the pa version of the page. Why no update was needed?

Norma said that pa is already ok

PR #1072, opened 2023-11-23
Cancel notification button shown as administrator role only

NicolaGiornetta:
## Short description
Cancel notification button is only shown when the user have administrator role.

- Added check via useHasPermissions hook
- Added new test

## How to test
Enter PA as operator. You shouldn't see cancel notification button in notification detail.

### DIFF
--- a/packages/pn-pa-webapp/src/components/Notifications/NotificationDetailTableSender.tsx
+++ b/packages/pn-pa-webapp/src/components/Notifications/NotificationDetailTableSender.tsx
@@ -11,10 +11,14 @@ import {
   dataRegex,
   formatDate,
   formatEurocentToCurrency,
+  useHasPermissions,
   useIsCancelled,
 } from '@pagopa-pn/pn-commons';
 import { Tag, TagGroup } from '@pagopa/mui-italia';
 
+import { PNRole } from '../../models/user';
+import { useAppSelector } from '../../redux/hooks';
+import { RootState } from '../../redux/store';
 import { TrackEventType } from '../../utility/events';
 import { trackEventByType } from '../../utility/mixpanel';
 import ConfirmCancellationDialog from './ConfirmCancellationDialog';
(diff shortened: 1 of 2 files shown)


packages/pn-pa-webapp/src/components/Notifications/NotificationDetailTableSender.tsx
```diff
@@ -46,7 +54,9 @@ const NotificationDetailTableSender: React.FC<Props> = ({ notification, onCancel
 
   const handleModalCloseAndProceed = () => {
     setShowModal(false);
-    onCancelNotification();
+    if (userHasAdminPermissions) {
```
clombardi, line 57: If my understanding is correct, this function shouldn't be called if `userHasAdminPermissions` is false (since in such scenario `ConfirmCancellationDialog` would never be rendered). Hence I find confusing to avoid the execution of just part of the code, it gives the impression that for an user without admin permission, the function could be executed and should just "close" the modal, which is not the actual behavior.
For that reason, I suggest to delete the conditional, and eventually add a comment about the function to indicate that it would never be called if `userHasAdminPermissions` is false, since in such a case the cancel procedure wouldn't even  be launched.

clombardi (changes requested):
Both the implementation and the added tests seem sound. I just ask for a slight change in the code that avoids what IMHO is a false impression about the behavior.

packages/pn-pa-webapp/src/components/Notifications/NotificationDetailTableSender.tsx
clombardi: We discussed this detail with @NicolaGiornetta, we have incompatible visions regarding this detail.
Nicola says that the conditional was added to further protection in case a hacker-ish (and evil) user wants to cancel a notification (to the benefit of some private and the damage to the state) even if he/she has not the needed grants, since the users can actually alter the behavior of the Web page, and **the BE operation lacks the role-related protection**.  
So I suggested him to leave the conditional, with a comment that prevents some other colleague to accidentally remove it in the future, since it has no evident sense. To this, Nicola replies that the user could even access to the GitHub repo (which is indeed public) and this fact could help him discover the way to fool SEND.  
IMHO for such an user would be easier to just sneak the API call to do and launch it from Postman / Insomnia / some similar tool, and in this situation I favor the readability of code over the possibility (I think remote) of hacking.

Since we don't agree in this point, I leave this point to @ZorbaDimatteo to decide.

clombardi (approved):
A decision was taken regarding the comment I had signaled, hence this PR is ready to commit.

PR #1073, opened 2023-11-23
PN-8797-fe-pg-nel-portale-pg-sostituire-lemail-di-assistenza-nellheader-con-il-link-assistenza

ImJimMoriarty:
## Short description
The behavior of the support link in PG has been modified for logged-in users.

- Added the link `${SELFCARE_BASE_URL}/assistenza` to `handleAssistanceClick()` once the user is logged in.
- Added on CDN config file `/assistenza`.

## How to test
Simply run `yarn start` for PG and check if the assistance link redirects to a different path after logging in.

### DIFF
--- a/packages/pn-personagiuridica-webapp/src/App.tsx
+++ b/packages/pn-personagiuridica-webapp/src/App.tsx
@@ -228,7 +228,9 @@ const ActualApp = () => {
   };
 
   const handleAssistanceClick = () => {
-    trackEventByType(TrackEventType.CUSTOMER_CARE_MAILTO, { source: 'postlogin' });
+    trackEventByType(TrackEventType.CUSTOMER_CARE_MAILTO, {
+      source: sessionToken ? 'postlogin' : 'prelogin',
+    });
     /* eslint-disable-next-line functional/immutable-data */
     window.location.href = sessionToken
       ? `${SELFCARE_BASE_URL}/assistenza`
--- a/packages/pn-personagiuridica-webapp/src/App.tsx
+++ b/packages/pn-personagiuridica-webapp/src/App.tsx
@@ -65,7 +65,7 @@ const App = () => {
 };
 
 const ActualApp = () => {
-  const { MIXPANEL_TOKEN, PAGOPA_HELP_EMAIL, VERSION } = getConfiguration();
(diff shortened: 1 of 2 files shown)


packages/pn-personagiuridica-webapp/src/App.tsx
```diff
@@ -230,9 +230,10 @@ const ActualApp = () => {
   const handleAssistanceClick = () => {
     trackEventByType(TrackEventType.CUSTOMER_CARE_MAILTO, { source: 'postlogin' });
```
clombardi, line 231: I wonder whether the TrackEventType shouldn't be different if the user is redirected to `/assistenza` instead of a mail sender.

clombardi (changes requested):
Hi, please check the right type for the `trackEvent`, since now it won't be sending a mail but redirecting if the user is logged.

packages/pn-personagiuridica-webapp/src/App.tsx
clombardi: @ZorbaDimatteo please verify that the TrackEventType is right even if the user is logged, thanks!

clombardi (commented):
Left a comment to @ZorbaDimatteo since there is a definition which is not clear, at least to me.

packages/pn-personagiuridica-webapp/src/App.tsx
clombardi: @ZorbaDimatteo if I understood you correctly, the source should be (say) `prelogin` or `postlogin` depending on whether there is a logged user when the link is pressed; I point this since this function is called in both scenarios, logged user or not. Am I right?

ZorbaDimatteo: @clombardi @ImJimMoriarty yes. When the function is called when user is not logged in, Assistenza button will have a mailto to assistance email and the tracking event would have source: "preLogin", otherwise, when user is not logged in, button assistenza will send the user to the new assistenza form and the event source will be "postLogin"

clombardi: OK, now the event' context can be obtained from the source.

clombardi (approved):
The only doubt I had has been resolved, ready to commit.

PR #1104, opened 2024-01-10
update pagopa/mui-italia version (pn-8759, pn-7135, pn-6687)

AndreaCimini90:
## Short description
Update pagopa/mui-italia version to fix some bugs

- Changed packages version
- Applied custom style to timeline components

## How to test
- Log with pa and check that current product is selected in the dropdown and that, in mobile version, timeline has right style
- Log with pf and check that, in mobile version, icons in the header aren't grey


packages/pn-commons/package.json
```diff
@@ -6,11 +6,11 @@
   "dependencies": {
     "@date-io/date-fns": "^2.15.0",
     "@mui/icons-material": "^5.14.3",
-    "@mui/lab": "5.0.0-alpha.140",
-    "@mui/material": "5.14.5",
+    "@mui/lab": "^5.0.0-alpha.140",
```
ZorbaDimatteo, line 9: @AndreaCimini90 why the use of ^ for these two packages?

AndreaCimini90: To use the same that are used in @pagopa/mui-italia

ZorbaDimatteo (approved):

PR #1105, opened 2024-01-10
fix(PN-9089): [FE] QRcode reading error message mitigation

SarahDonvito:
## Short description
Updated AccessDenied component to show message when QRcode is read by the wrong receiver or without proxies.

- Added icon props to make the AccessDenied component behave differently in this case
- Passed icon prop as <IllusQuestion /> to AARGuard in PF and PG
- Created IllusQuestion to import it in AARGuard component
- Moved setting item method from AARGuard in RouteGuard in PG

- Updated README.md with new command for running single file test

## How to test
Open the notification link via qrcode of an aar (both for PG and for PF); check if the error message (icon, title, subtitle and button "vai alla homepage") is shown.


packages/pn-commons/src/pages/AccessDenied.tsx
```diff
@@ -16,6 +17,7 @@ const AccessDenied: React.FC<Props> = ({
   goToHomePage,
   message,
   subtitle,
+  qrError = false
```
AndreaCimini90, line 20: This is a common component, so i wouldn't use a so specific property.
I would change this property with a more generic "icon" property. This property will be of type Illustration (check the right type).
So IllusQuestion will not used here but passed from outside

packages/pn-commons/src/pages/AccessDenied.tsx
```diff
@@ -31,17 +33,21 @@ const AccessDenied: React.FC<Props> = ({
     <Stack
       direction="column"
       alignItems="center"
-      my={4}
+      my={6}
```
AndreaCimini90, line 36: Here and down, we override the original style. This change also the style of the AccessDenied used when user is not logged in.
That It is ok?

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1114, opened 2024-01-17
chore: enabled mixpanel tracking for portale login

ZorbaDimatteo:
## Short description
Re-enabled mixpanel tracking for portale login


ZorbaDimatteo:
After runing refresh:monorepo , yarn.lock has been updated. Is it ok?

PR #1118, opened 2024-01-22
PN-9541 - Logout from ToS and PP page

alessandrogelmi:
## Short description
As decided, logging out from the ToS and PP pages should not redirect the user to the login page but instead keep them on those pages as non-logged in

- Edited handleUserLogout function

## How to test
On PF/PG try to go on PP or ToS page and click on logout button. You should remain on that page

### DIFF
--- a/packages/pn-personafisica-webapp/src/App.tsx
+++ b/packages/pn-personafisica-webapp/src/App.tsx
@@ -236,8 +236,12 @@ const ActualApp = () => {
   });
 
   const handleUserLogout = () => {
+    const isPpOrTosPage =
+      path[1] === routes.PRIVACY_POLICY.split('/')[1] ||
+      path[1] === routes.TERMS_OF_SERVICE.split('/')[1];
+
     void dispatch(logout());
-    goToLoginPortal();
+    !isPpOrTosPage && goToLoginPortal();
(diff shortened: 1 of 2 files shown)


packages/pn-personafisica-webapp/src/App.tsx
```diff
@@ -236,8 +236,12 @@ const ActualApp = () => {
   });
 
   const handleUserLogout = () => {
+    const isPpOrTosPage =
```
ZorbaDimatteo, line 239: @alessandrogelmi let's try to generalize this behaviour for all public pages, we can make al list or a map containing public routes related to the single webapp and an utility method (e.g. isPublicPage) that returns true o false

ZorbaDimatteo (changes requested):
@alessandrogelmi let's try to generalize this approach to all public pages

packages/pn-personafisica-webapp/src/App.tsx
alessandrogelmi: After an analysis, I discovered that removing the forced login redirect (goToLoginPortal function) resolves everything. Logging out from pages that are not within the SessionGuard automatically stays on that page.

PR #1120, opened 2024-01-23
fix/PN-9145: fix on catching error for duplicated protocol number from FE

SarahDonvito:
## Short description
Fix on createNewNotification in case of error for duplicated protocol number when creating a notification.

- fixed createNotification method to catch error for duplicated protocol number in this part too (the error was already managed showing the toast of error)
- implemented unit test "notification failed for duplicated protocol number"

## How to test
Login in as PA, click on "invia una notifica", create a notification with a protocol number. Try to create a new notification with the same protocol number of the previous one. Check a toast of error appear.


packages/pn-pa-webapp/src/pages/__test__/NewNotification.page.test.tsx
```diff
@@ -261,6 +263,69 @@ describe('NewNotification Page without payment', async () => {
...
+    fireEvent.click(buttonSubmit);
+    await waitFor(() => {
+      expect(mock.history.post).toHaveLength(1);
+    });
+    const finalStep = result.queryByTestId('finalStep');
+    expect(finalStep).not.toBeInTheDocument();
+
+    // check if toast is in the document
+    const snackBar = await waitFor(() => result.getByTestId('snackBarContainer'));
+    expect(snackBar).toBeInTheDocument();
```
AndreaCimini90, line 327: check also that the toast has the right message

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1123, opened 2024-01-26
fix(PN-9412): fix alert message when AAR not available for PA

SarahDonvito:
## Short description
implemented alert when document aar is not immediately available after 120 days

- Added TimedMessage component in AAR box in order to alert aar is not immediately available

## How to test
Login as PA, open a notification sent 120 days ago from _today_ and try to download the aar document from aar box. An alert will appear in the same section and not at the top of the timeline.


packages/pn-pa-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -24,6 +24,7 @@ import {
   useIsCancelled,
   useIsMobile,
 } from '@pagopa-pn/pn-commons';
+import { dateIsLessThan10Years } from '@pagopa-pn/pn-commons/src/utility/date.utility';
```
ZorbaDimatteo, line 27: @SarahDonvito you should export the new function in the index file of pn-commons, and then using it as anything else imported by pn-commons (eg. see the line above this comment)

packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
ZorbaDimatteo, line 1: @SarahDonvito I suppose that the check dateIsLessThan10Years should be used also in pf/pg when using NotificationDetailDocuments component, see line 551 (ofc together with any other condition)

ZorbaDimatteo (changes requested):

ZorbaDimatteo (approved):

PR #1127, opened 2024-02-07
chore: fixes on payment caching

alessandrogelmi:
## Short description
Added `eventTimestamp` to paymentCache validation schema. This property came from timeline.

- Changed validation schema
- Changed return statement on redux 

## How to test
When opening a notification you should't see warning about cache validation. 
To test the return, you should pay a notice and when you return on SEND you should see the notice updated

### DIFF
--- a/packages/pn-commons/src/models/NotificationDetail.ts
+++ b/packages/pn-commons/src/models/NotificationDetail.ts
@@ -126,6 +126,7 @@ export interface PrepareAnalogDomicileFailureDetails extends BaseDetails {
 
 interface BaseDetails {
   recIndex?: number;
+  eventTimestamp?: string;
 }
 
 interface DelegateInfo {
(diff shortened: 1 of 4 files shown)


packages/pn-commons/src/models/NotificationDetail.ts
```diff
@@ -219,6 +219,7 @@ export interface PaidDetails extends BaseDetails {
   creditorTaxId: string;
   noticeCode: string;
   uncertainPaymentDate?: boolean;
+  eventTimestamp?: string;
```
AndreaCimini90, line 222: Check the open api on delivery. This field maybe is for all the details and can be moved into BaseDetails

AndreaCimini90 (changes requested):

packages/pn-commons/src/models/NotificationDetail.ts
alessandrogelmi: Ok, I moved it into BaseDetails because it is also used in `NotificationViewedCreationRequestDetailsV23`, `NotificationRADDRetrievedDetails`, `RefinementDetailsV23`, `SendDigitalProgressDetailsV23` and `NotificationPaidDetailsV23`

ZorbaDimatteo (approved):

AndreaCimini90 (approved):

PR #1130, opened 2024-02-09
fix: fix in products.json

ZorbaDimatteo:


AndreaCimini90 (approved):

AndreaCimini90 (changes requested):
The problem is on the login section of the pf webapp. We don't configure the products list, but mui-italia download the list from https://selfcare.pagopa.it/assets/products.json, and this file contains the wrong url.
The products.json located in pg webapp, I think it is never used

ZorbaDimatteo:
@AndreaCimini90 yes, I'm aware of that. In fact the Jira issue correctly report that. Anyway, the product.json is used or is a backup only for some scenarios regarding pg (in particular: it is possible that in certain path under selfcare cdn our csp blocks the downloading of certain files). 
I think this pr should be approved without any changes

AndreaCimini90 (approved):

PR #1145, opened 2024-02-22
Support page - Feat/pn 10040

AndreaCimini90:
## Short description
Added support page for those users that are logged in.

- Added SupportPage component
- Added assistenza path to the routes
- Updated the behaviour of the support link (in the header) to manage the user logged case
- Created a utility file to manage the different app layouts
- Added new localization file
- Added new configuration for the support privacy policy link
- Created validator for the support page
- Added new test cases

## How to test
- got to pf portal as a not logged in user -> click on support link -> check that email provider is opened
- go to pf portal as logged in user -> click on support link -> check that new support page is shown -> submit the form and check that support form is opened


packages/pn-personafisica-webapp/src/pages/Support.page.tsx
```diff
@@ -0,0 +1,216 @@
...
+        })
+        .catch(() => {});
+    }
+  };
+
+  return (
+    <Prompt
+      title={t('exit-title')}
+      message={t('exit-message')}
+      eventTrackingCallbackPromptOpened={() => {}} // impostare eventi tracking previsti
```
ZorbaDimatteo, line 123: @AndreaCimini90 probably it's better to refactor PromptComponent and make these eventTracking properties optional?

packages/pn-personafisica-webapp/src/pages/Support.page.tsx
```diff
@@ -0,0 +1,216 @@
...
+      eventTrackingCallbackCancel={() => {}} // impostare eventi tracking previsti
+      eventTrackingCallbackConfirm={() => {}} // impostare eventi tracking previsti
+    >
+      <Grid container justifyContent="center" alignContent="center">
```
ZorbaDimatteo, line 127: @AndreaCimini90 I'd prefer to use a Stack over grid layout for one-dimensional layout

packages/pn-personafisica-webapp/src/pages/Support.page.tsx
AndreaCimini90: @ZorbaDimatteo done

packages/pn-personafisica-webapp/src/pages/Support.page.tsx
AndreaCimini90: @ZorbaDimatteo done

ZorbaDimatteo (approved):

PR #1158, opened 2024-03-14
fix(PN-10222): fix error message management

SarahDonvito:
## Short description
fix error message management in pf-login and add icon and cta in errorpage

- Add icon in Dialog in LoginError
- Add cta button to go back to login
- Fix message making it related to the error received

## How to test
Go on with the login in step and cancel the flow in the spid credentials. Check the error message is consistent with the error code (you can see it in the url)

### DIFF
--- a/packages/pn-personafisica-login/src/pages/loginError/LoginError.tsx
+++ b/packages/pn-personafisica-login/src/pages/loginError/LoginError.tsx
@@ -86,13 +86,8 @@ const LoginError = () => {
   }, []);
 
   return (
-    <Dialog
-      fullScreen={true}
-      open={true}
-      aria-labelledby="dialog-per-messaggi-di-errore"
-      id="errorDialog"
-    >
-      <Box m="auto" sx={{ textAlign: 'center', width: '100%' }}>
+    <Dialog fullScreen={true} open={true} aria-labelledby="dialog-per-messaggi-di-errore">
+      <Box m="auto" sx={{ textAlign: 'center', width: '100%' }} id="errorDialog">
         <IllusError />
         <Typography variant="h5" sx={{ fontSize: '18px', fontWeight: '600' }} mt={5}>
           {title}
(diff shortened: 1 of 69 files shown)


packages/pn-personafisica-login/src/pages/loginError/LoginError.tsx
```diff
@@ -54,13 +56,14 @@ const LoginError = () => {
...
+        return t('loginError.message');
     }
   };
 
+  const goToLogin = () => navigate(ROUTE_LOGIN);
   const title = t('loginError.title');
-  const message = (
-    <Trans i18nKey="loginError.message" ns="login">
+  const errorMessage = (
+    <Trans i18nKey={getErrorMessage()} ns="login">
```
AndreaCimini90, line 66: I'm not sure about this. The i18nKey wants the path to the label not the label itself.
For example 'loginError.message' and not t('loginError.message').

Furthermore {getErrorMessage()} is useless if we use the i18nKey.

We can remove the trans component because, in this case, is useless and use getErrorMessage() below, in the row where there is {errorMessage}

packages/pn-personafisica-login/src/pages/loginError/__test__/LoginError.test.tsx
```diff
@@ -55,7 +55,7 @@ describe('LoginError component', () => {
     const message = getById(errorDialog, 'message');
     expect(message).toHaveTextContent('loginError.message');
     // Wait login redirect
-    await new Promise((t) => setTimeout(t, 200));
+    await new Promise((t) => setTimeout(t, 2000));
```
AndreaCimini90, line 58: we can use fake timers, to prevent tests from taking too long. here and below

AndreaCimini90 (changes requested):

packages/pn-personafisica-login/src/pages/loginError/__test__/LoginError.test.tsx
```diff
@@ -44,6 +38,7 @@ describe('LoginError component', () => {
   });
 
   it('login technical error - code generic', async () => {
+    vi.useFakeTimers();
```
AndreaCimini90, line 41: At the end of the test case, add vi.useRealTimers(); Do it for all test cases

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1176, opened 2024-04-03
Get TraceId from header - Fix/pn 9848

AndreaCimini90:
## Short description
For those errors sent by the ApiGateway, we must take the TraceId from the header.

- Get the traceId from the headers if the one in the body is undefined - AppResponse.ts
- Removed all occurrences of `performThunkAction` to ensure that all requests use `parseError`

## How to test

- Login with PF or PG
- With Netify mutate the `https://webapi.dev.notifichedigitali.it/bff/v1/mandate` API request with this body
```
{
    "delegate": {
        "fiscalCode": "csrggl44l13h501e",
        "person": true,
        "displayName": "Gaio Giulio Cesare",
        "firstName": "Gaio Giulio",
        "lastName": "Cesare"
    },
    "visibilityIds": [],
    "verificationCode": "40762",
    "dateto": "2024-10-25"
}
```
- Check the Mixpanel Event `SEND_TOAST_ERROR` contains the property `traceId`

Since we have removed all occurrences of `performThunkAction` (from `exchangeToken` and Zendesk support api), try logging in (with PF, PG, and PA) to verify that `exchangeToken` works as expected. Additionally, on PF, attempt to make a Zendesk request to ensure it works as expected


AndreaCimini90 (commented):
parseError cannot be the right place because we don't use it everywhere. The createAppResponseError should be the right place

alessandrogelmi (approved):

PR #1179, opened 2024-04-11
feat(PN-10563): remove jest-axe file tests in PA

SarahDonvito:


AndreaCimini90 (changes requested):
Pr is ok. Check for tests in pn-commons also and the packages into the package.json and their use.

TIP: search for jest-axe

ZorbaDimatteo (changes requested):
@SarahDonvito the pr it's ok, but remove also these two devDependencies
```
"@types/jest-axe": "^3.5.3",
"jest-axe": "^6.0.0",
```

ZorbaDimatteo (approved):

PR #1205, opened 2024-05-10
Feature/pn-10866 - Add layer for statistics data transformation (PA)

mflauti:
## Short description
This PR adds the utility/StatisticsData module inside the pa-webapp to achieve the goal of performing data transformation from the dto received from the be to a collection of data structures which are ready to be used by the components which form the sender dashboard.

- add StatisticsData abstract class, which act as superclass for every concrete StatisticsData
- add a collection of specialized StatisticsData classes. Any single class implements the logic to generate the transformed data for a particular component to be presented on the sender dashboard:
  - FiledStatisticsData
  - LastStateStatisticsData
  - DeliveryModeStatisticsData
  - DigitalStateStatisticsData
  - DigitalMeanTimeStatisticsData
  - DigitalErrorsDetailStatisticsData
  - DigitalAttemptsStatisticsData
- add StatisticsDataFactory which is responsible for instantiating objects having type StatisticsData and use them to build the transformed data
- add StatisticsDataFactoryManager which is responsible for creating, storing and managing a factory instance

## How to test
Go to Statistics page and verify the statistics slice of the redux store got updated with proper data
Be sure to set the IS_STATISTICS_ENABLED environment variable to true in conf/config.json


packages/pn-pa-webapp/src/models/Statistics.ts
```diff
@@ -0,0 +1,162 @@
...
+  DELIVERING = 'DELIVERING',
+  DELIVERED = 'DELIVERED',
+  VIEWED = 'VIEWED',
+  EFFECTIVE_DATE = 'EFFECTIVE_DATE',
+  UNREACHABLE = 'UNREACHABLE',
+  CANCELLED = 'CANCELLED',
+  REFUSED = 'REFUSED',
+}
+
+export enum DeliveryMode { // export from pn-commons/../NotificationDetail.ts???
```
ZorbaDimatteo, line 12: @maurizioflauti to address this comment: no, they should not be exported from NotificationDetail types

ZorbaDimatteo (approved):

packages/pn-pa-webapp/src/models/Statistics.ts
mflauti: ok

packages/pn-pa-webapp/src/models/Statistics.ts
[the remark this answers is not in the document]
AndreaCimini90: When we will integrate the bff api, the api layer will be removed together with routes.
For the models, we can decide if remove them or not. For the others api, we aren't removing them due to the discussion about the front-end mapper.
I am of the opinion that, if the models aren't in pn-commons, they can be removed

PR #1216, opened 2024-05-22
PN-10930-fe-pa-notifica-inviata-digitalmente-nonostante-domicilio-digitale-sia-stato-inserito-ma-non-selezionato-in-fase-di-invio-notifica

ImJimMoriarty:
## Short description
This pull request addresses an issue where the `digitalDomicile` field value was not being cleared when the `showDigitalDomicile` checkbox wasn't toggled. The update ensures that the field value is set to an empty string whenever this statment is false.

- clear the `digitalDomicile` field value by setting it to an empty string ("") when the `showDigitalDomicile` checkbox is toggled.

## How to test
Start PA and navigate to the related component (Recipient). Try to replicate the behavior described in the [ticket](https://pagopa.atlassian.net/browse/PN-10930?atlOrigin=eyJpIjoiZTE1MmRiOWViZTk2NDVhOWEyMTgxMTIwNzJmNzg4ZWEiLCJwIjoiaiJ9).


packages/pn-pa-webapp/src/components/NewNotification/Recipient.tsx
```diff
@@ -541,6 +541,9 @@ const Recipient: React.FC<Props> = ({
                             `recipients[${index}].showDigitalDomicile`,
                             !values.recipients[index].showDigitalDomicile
                           );
+                          if (!values.recipients[index].showDigitalDomicile) {
```
AndreaCimini90, line 544: This fix leads a bug: enable the digital address -> continue button is disabled (right) -> fill the input field -> continue button is enabled (right) -> disable digital address -> continue button is still enabled (right) -> enable digital address -> the field is empty (right) and the continue button is still enabled (WRONG)

Having two setFieldValue breaks the validation, maybe because, when the second setFieldValue is run, the first has not finished yet.

To solve the bug we can use one only setFieldValue for both the checkbox and the input value. To do this we can act on `recipients[${index}]` and override showDigitalDomicile and digitalDomicile instead of `recipients[${index}].showDigitalDomicile` and `recipients[${index}].digitalDomicile`.

TIPS: we can use the spread operator

the `if (!values.recipients[index].showDigitalDomicile)` can be avoided because, even if we enable or disable the digital address, it's ok to set the digitalDomicile to empty.

This solves the bug, but there is something that must be still fixed. When we enable the digital address the second time, the input is touched and the required error is shown.

So before setting the value (before the setFieldValue), reset the touched status with `setFieldTouched`. This function accepts three parameters: the field (`recipients[${index}].digitalDomicile`), the touched status (false) and the shouldValidate (false because the validation will be run by the setFieldValue)

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1219, opened 2024-05-23
PN-10641-fe-pf-pg-recapiti-messaggio-di-errore-inserimento-codice-otp-non-mostrato-al-secondo-tentativo

ImJimMoriarty:
## Short description
This pull request improves error handling and subscription management by following the same schema of PG and the [FIGMA](https://www.figma.com/design/O6q8HrX3opuykOcj9idcKT/PF-Destinatari?node-id=274-72592#757588774).

- updated subscription and unsubscription of error events using `AppResponsePublisher` in `useEffect` hook.

## How to test
Start PF and navigate into contacts page and attempt to add a new contact using any available method, during the contact addition process, type an incorrect OTP code and repeat it two to three times consecutively, then observe the behavior of the application.


AndreaCimini90 (changes requested):

AndreaCimini90 (dismissed):

AndreaCimini90:
@ImJimMoriarty the bug is still present. The problem is the value of the property ValidationCodeMaxAttempts. During our tests, the value was 2 and this prevented the bug from appearing. With a ninja deploy with ValidationCodeMaxAttempts set to 3, the bug appears.

The problem comes from the use of an external state setErrorMessage (**external error**), and an internal state setInternalError (**internal error**). The internal error reacts to the changes of external error, but between attempt 1 and 2 if we change the code (this is important to make the bug appear) something happens:
- the internal error is reset
- the external error doesn't change (because the error returned by the be is the same)
in this way no changes is triggered and the error is not shown until a further attempt is made (because the error returned by the be changes)

I suggest to try to remove the external error state and use ref and useImperative to export from the inner component the function that updates the internal error.

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1231, opened 2024-06-03
Fix/PN-11403 - Manage api error for sender dashboard

mflauti:
## Short description
Handle the case the server responds with a 4xx or 5xx response to a getStatistics api request using the ApiErrorWrapper

- use ApiErrorWrapper inside Statistics.page

## How to test
Prerequisite: IS_STATISTICS_ENABLED env var should be set to true
- In Statistics.page change the cxId at line 70 with any random string to simulate a bad request and force the server to respond with a 400.


AndreaCimini90 (changes requested):
Add a test case into Statistics.page.test.tsx to check the error handling.

TIPS: Check ApiKeys.page.test.tsx to see how to do

AndreaCimini90 (approved):

AndreaCimini90 (approved):

PR #1248, opened 2024-06-07
PN-11527-fe-e-possibile-generare-unapi-key-inserendo-solo-uno-spazio-nel-campo-nome

ImJimMoriarty:
## Short description
This pull request fixes the minimum character validation for the name field in the NewApiKey form. Previously, the form allowed names with less than 3 characters, which is not desirable. With this fix, the form will now correctly validate that the name has at least 3 characters before allowing submission.

- added a minimum of 3 characters to the name field in the validation schema
- updated related tests
- added translations for each language

## How to test
Start PA and try to create an api key with less than 3 character and see if the error appear.

### DIFF
--- a/packages/pn-pa-webapp/public/locales/de/apikeys.json
+++ b/packages/pn-pa-webapp/public/locales/de/apikeys.json
@@ -80,6 +80,7 @@
     "form-label-name": "Gib deinem API-Schlüssel einen Namen",
     "form-placeholder-name": "Bitte einen Namen eingeben",
     "form-error-name": "Definiere einen Namen für den API-Schlüssel",
+    "form-error-minimum": "Der Name muss mindestens 3 Zeichen lang sein",
     "form-label-groups": "Wähle die Gruppen aus, denen der API-Schlüssel zugewiesen werden soll",
     "form-placeholder-groups": "Gruppe suchen",
     "form-error-groups": "Mindestens eine Gruppe auswählen",
--- a/packages/pn-pa-webapp/public/locales/de/apikeys.json
+++ b/packages/pn-pa-webapp/public/locales/de/apikeys.json
@@ -80,7 +80,6 @@
     "form-label-name": "Gib deinem API-Schlüssel einen Namen",
     "form-placeholder-name": "Bitte einen Namen eingeben",
     "form-error-name": "Definiere einen Namen für den API-Schlüssel",
-    "form-error-minimum": "Der Name muss mindestens 3 Zeichen lang sein",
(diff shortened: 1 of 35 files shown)


packages/pn-pa-webapp/src/pages/NewApiKey.page.tsx
```diff
@@ -46,7 +46,7 @@ const NewApiKey = () => {
   });
 
   const validationSchema = yup.object({
-    name: yup.string().required(tkp('form-error-name')),
+    name: yup.string().min(3, tkp('form-error-minimum')).required(tkp('form-error-name')),
```
AndreaCimini90, line 49: Why the min set to 3? Is there some specification in the openapi definition? 
If not, replace the `.min(3, tkp('form-error-minimum'))` with `.matches(dataRegex.noSpaceAtEdges, t('no-spaces-at-edges'))`.
If yes, the fix not solves the bug because we can add 3 or more spaces and the error will disappear. In this case we have to add the `.matches(dataRegex.noSpaceAtEdges, t('no-spaces-at-edges'))` in addition to the `.min(3, tkp('form-error-minimum'))`

packages/pn-pa-webapp/public/locales/de/apikeys.json
```diff
@@ -80,6 +80,7 @@
     "form-label-name": "Gib deinem API-Schlüssel einen Namen",
     "form-placeholder-name": "Bitte einen Namen eingeben",
     "form-error-name": "Definiere einen Namen für den API-Schlüssel",
+    "form-error-minimum": "Der Name muss mindestens 3 Zeichen lang sein",
```
AndreaCimini90, line 83: remove the custom localization here and below

packages/pn-pa-webapp/src/pages/__test__/NewApiKey.page.test.tsx
```diff
@@ -72,6 +72,11 @@ describe('NewApiKey component', async () => {
     expect(submitButton).toBeEnabled();
     expect(result.container).not.toHaveTextContent(/form-error-name/);
 
+    // fill api key name with less than 3 characters: invalid form, submit is disabled, error message shown
+    await testInput(form, 'name', 'ab');
```
AndreaCimini90, line 76: Update this check with a string with only spaces

AndreaCimini90 (changes requested):

AndreaCimini90 (dismissed):

AndreaCimini90 (approved):

PR #1254, opened 2024-06-12
feat/PN-11564 - Whole page component to manage the empty state

mflauti:
## Short description
Shows a Whole page component to manage the empty state when filtering by a certain range of dates returns no data

- add a boolean 'isEmpty' field to StatisticsParsedResponse model to immediately signal there's no significant response data
- change Statistics.page to show the whole page component if isEmpty is true
- change the text showed as description when filtering by dates returns an empty result

## How to test
- Go to Sender Dashboard, filter using a narrow dates range to force an empty data response and verify the behaviour is as described by the linked task

### DIFF
--- a/packages/pn-pa-webapp/src/redux/statistics/actions.ts
+++ b/packages/pn-pa-webapp/src/redux/statistics/actions.ts
@@ -5,7 +5,6 @@ import {
   parseError,
 } from '@pagopa-pn/pn-commons';
 import { createAsyncThunk } from '@reduxjs/toolkit';
-import { isEmpty } from 'lodash';
 
 import { apiClient } from '../../api/apiClients';
 import { SenderDashboardApiFactory } from '../../generated-client/sender-dashboard';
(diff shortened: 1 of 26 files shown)


packages/pn-pa-webapp/src/redux/statistics/actions.ts
```diff
@@ -44,6 +47,7 @@ export const getStatistics = createAsyncThunk(
             lastDate: response.data.lastDate,
             startDate: response.data.startDate,
             endDate: response.data.endDate,
+            isEmpty: isResponseEmpty(response.data as StatisticsResponse),
```
AndreaCimini90, line 50: Avoid to add an extra field isEmpty. Try to use the redux selector technique https://redux.js.org/usage/deriving-data-selectors

AndreaCimini90 (changes requested):

packages/pn-pa-webapp/src/redux/statistics/actions.ts
mflauti: ok

AndreaCimini90 (approved):

AndreaCimini90 (approved):

PR #1256, opened 2024-06-13
PN-7276 - Add required rule on pn-validator

alessandrogelmi:
## Short description
Add required rule on pn-validator

- Add rule
- Add tests of rule

### DIFF
--- a/packages/pn-validator/src/__test__/ValidatorBuilder.test.ts
+++ b/packages/pn-validator/src/__test__/ValidatorBuilder.test.ts
@@ -103,28 +103,4 @@ describe('Test ValidatorBuilder', () => {
     const results = dummyValidatorBuilder.validate('no match', {});
     expect(results).toBe('Custom error message');
   });
-
-  it('isRequired - check if validate works (value not required)', () => {
-    const dummyValidatorBuilder = new ValidatorBuilder<any, any>();
-    const rules = dummyValidatorBuilder.getTypeRules();
-    rules.isString().isRequired();
-    const results = dummyValidatorBuilder.validate('Value', {});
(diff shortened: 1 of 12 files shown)


packages/pn-validator/src/ruleValidators/__test__/ArrayRuleValidator.test.ts
```diff
@@ -1,21 +1,29 @@
...
 vi.mock('../../rules/ForEachElement', () => {
-    return {
-        ForEachElement: vi.fn()
-    }
+  return {
+    ForEachElement: vi.fn(),
+  };
+});
+
+vi.mock('../../rules/Required', () => {
```
AndreaCimini90, line 23: Remove the test about the required rule and leave only that is in the common rule validator. Do it for all tests.

After that add the check into the ValidatorBuilder.test.tsx

AndreaCimini90 (changes requested):

packages/pn-validator/src/__test__/ValidatorBuilder.test.ts
```diff
@@ -101,4 +103,28 @@ describe('Test ValidatorBuilder', () => {
     const results = dummyValidatorBuilder.validate('no match', {});
     expect(results).toBe('Custom error message');
   });
+
+  it('isRequired - check if validate works (value not required)', () => {
```
AndreaCimini90, line 107: You can remove these three tests. The required rule is tested into the RequireRule test and the custom message is already tested above

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1281, opened 2024-07-23
PN-12148 - Use selected language on every page of pf-login

alessandrogelmi:
## Short description
Load lang for every page on pf-login

- Moved initial set of language on App.tsx 

## How to test
Start pf-login, try to change language and then navigate to `auth/login/error`. You should see the the error message translated


packages/pn-personafisica-login/src/App.tsx
```diff
@@ -2,29 +2,40 @@ import { useEffect, useMemo, useState } from 'react';
...
     callback: () => console.log(`v${configuration.VERSION}`),
   });
 
-  const { t } = useTranslation(['common']);
+  const { t, i18n } = useTranslation(['common']);
   const [isInitialized, setIsInitialized] = useState(false);
 
+  const handleSetUserLanguage = async () => {
+    const language = getSessionLanguage() || 'it';
+    await i18n.changeLanguage(language);
```
AndreaCimini90, line 30: This can lead to an extra render. We can try to create an i18next language detector plugin https://www.i18next.com/misc/creating-own-plugins.

If it'll work, we can do the same thing for all the apps

packages/pn-personafisica-login/src/pages/loginError/LoginError.tsx
```diff
@@ -94,7 +93,7 @@ const LoginError = () => {
           {getErrorMessage()}
         </Typography>
         <Button id="login-button" variant="contained" onClick={goToLogin}>
-          {getLocalizedOrDefaultLabel('common', 'button.go-to-login', 'Accedi')}
+          {t('loginPage.loginBox.title')}
```
AndreaCimini90, line 96: Why do we change button.go-to-login into loginPage.loginBox.title?

AndreaCimini90 (changes requested):

packages/pn-personafisica-login/src/pages/loginError/LoginError.tsx
alessandrogelmi: Ok, I've restored button.go-to-login and added the translations for every language.
@ZorbaDimatteo , I found the translations on pn-pa-webapp.

AndreaCimini90 (approved):

AndreaCimini90:
@ZorbaDimatteo @alessandrogelmi the pr is ok. Before merging i have one question.
This is a way to detect also the browser language?

alessandrogelmi:
I think we need to add `navigator` to the `order` array in the i18n config file (`order: ['querystring', 'sessionStorage']`).
Is it right @ZorbaDimatteo?

ZorbaDimatteo:
@AndreaCimini90 @alessandrogelmi
language detector from browser language will be implemented in a next feature/release (see "Bilinguismo amministrativo") 
I would not change requisite for this bug.
To me the pr is approved

ZorbaDimatteo (approved):

PR #1285, opened 2024-07-26
PN-11927 - Rework copy of not enough data on sender dashboard

alessandrogelmi:
## Short description
Rework copy of not enough data on sender dashboard

- Use Trans component in order to use Link inside copy

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.


packages/pn-pa-webapp/src/components/Statistics/EmptyStatistics.tsx
```diff
@@ -1,18 +1,59 @@
...
+type LinkDashboardProps = {
+  children?: React.ReactNode;
+};
+
+const LinkDashboard: React.FC<LinkDashboardProps> = ({ children }) => {
+  const { t } = useTranslation(['statistics']);
+  const navigate = useNavigate();
+
+  return (
+    <Link
```
AndreaCimini90, line 25: The style of the link doesn't is the the same of the figma. Maybe we have to use the ButtonNaked component. Check other links around the application

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1295, opened 2024-08-07
Feat/pn 11970

AndreaCimini90:
## Short description
Removed context from legal contacts

- Created two new dialog
- Copied code from context to PecContactItem
- Removed context usage from PecContactItem
- Reworked tests

## How to test
- Login with pf/pg -> go to "Recapiti" -> add/modify/delete "default" legal contact and check the everything is ok


packages/pn-personafisica-webapp/src/components/Contacts/PecContactItem.tsx
alessandrogelmi, line 1: 1. Seems that on large screen the text input not fit all width (as in TEST environment)
2. When I add a new PEC and then click the cancel button in the code modal, the input value is cleared.

packages/pn-personafisica-webapp/src/components/Contacts/PecContactItem.tsx
```diff
@@ -58,18 +85,53 @@ const PecContactItem = ({ value, verifyingAddress, blockDelete }: Props) => {
...
+        if (res.pecValid) {
+          // show success message
+          dispatch(
+            appStateActions.addSuccess({
+              title: '',
+              message: t(`legal-contacts.pec-added-successfully`, { ns: 'recapiti' }),
+            })
+          );
+          setModalOpen(null);
+          digitalElemRef.current.toggleEdit();
```
alessandrogelmi, line 123: When I add a new PEC, the field remains in edit mode even if the operation is successful.
However, when I edit an existing PEC, the behaviour is correct.

alessandrogelmi (changes requested):

packages/pn-personafisica-webapp/src/components/Contacts/LegalContacts.tsx
```diff
@@ -33,7 +33,7 @@ const LegalContacts = ({ legalAddresses }: Props) => {
       subtitle={t('legal-contacts.description', { ns: 'recapiti' })}
       avatar={<IllusEmailValidation />}
     >
-      <Box sx={{ width: { xs: '100%', lg: '50%' } }} data-testid="legalContacts">
+      <Box sx={{ width: '100%' }} data-testid="legalContacts">
```
alessandrogelmi, line 36: Do the same thing on PG

alessandrogelmi (changes requested):

PR #1309, opened 2024-08-26
modify directive frame-ancestor

leleOFA:
## Short description
In this pr change is required on the cdn for SelfCareHeaderPolicy and DefaultHeaderPolicy in the Content Security section on the frameAncestor value.

set ancestors null


ZorbaDimatteo (commented):
@leleOFA pr is ok, but try to follow pr template (adding information or removing section) and set also in the title of the pr the number of the relative issue/task, to help reviewers to understand better the context

ZorbaDimatteo (approved):

PR #1321, opened 2024-09-19
PN-12540 - Digital domicile mixpanel events

alessandrogelmi:
## Short description
Add mixpanel events for DomD

- Add some mixpanel events 

## How to test
Follow the flow described in Figma and verify that Mixpanel events are triggered


packages/pn-personafisica-webapp/src/redux/contact/reducers.ts
```diff
@@ -16,6 +22,7 @@ const initialState = {
   loading: false,
   digitalAddresses: [] as Array<DigitalAddress>,
   parties: [] as Array<Party>,
+  fromExternalInfo: {} as FromExternalInfo,
```
AndreaCimini90, line 25: use null

PR #1323, opened 2024-09-20
Pec addtion when sercq send is enabled as default address - pn-12537

AndreaCimini90:
## Short description
Manage pec addition when sercq send is enabled as default address

- Updated LegalContacts component
- Updated PecContactItem
- Updated SercqContacItem
- Updated tests

## How to test
Login with pf/pg -> enable sercq send as default address -> check that the view change accorndly -> add pec -> check that everything is ok


packages/pn-personafisica-webapp/src/components/Contacts/LegalContacts.tsx
alessandrogelmi, line 1: If PEC is enabled, it should be shown as the first element and divider should be hidden

alessandrogelmi (changes requested):

alessandrogelmi (approved):

PR #1341, opened 2024-10-04
feat(PN-12538): fix banner condition

SarahDonvito:
## Short description
fix banner condition PF

- Removed banner in Profile page
- Removed banner if notification is not viewed

## How to test

1. Login as PF
2. Go to a not viewed notification
3. Check banner is not present
4. Go to a viewed notification
5. Check banner is present

### DIFF
--- a/packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
+++ b/packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
@@ -19,6 +19,7 @@ import {
   NotificationDocumentType,
   NotificationPaymentRecipient,
   NotificationRelatedDowntimes,
+  NotificationStatus,
   PaymentAttachmentSName,
   PaymentDetails,
   PnBreadcrumb,
@@ -89,6 +90,7 @@ const NotificationDetail: React.FC = () => {
   );
 
   const isCancelled = useIsCancelled({ notification });
+  const isViewed = notification.notificationStatus === NotificationStatus.VIEWED;
   const currentRecipient = notification?.currentRecipient;
 
   const userPayments = useAppSelector((state: RootState) => state.notificationState.paymentsData);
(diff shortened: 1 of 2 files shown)


packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -89,6 +90,7 @@ const NotificationDetail: React.FC = () => {
   );
 
   const isCancelled = useIsCancelled({ notification });
+  const isViewed = notification.notificationStatus === NotificationStatus.VIEWED;
```
AndreaCimini90, line 93: This is not totally correct. We have to check if notification has been viewed or not. This check must be valid for all those status after viewed. With this check we are checking that the "current" status is Viewed but we are not considering other statuses after that.

TIPS: check how was done in the pg app

packages/pn-personafisica-webapp/src/pages/Profile.page.tsx
```diff
@@ -1,49 +1,24 @@
...
   useEffect(() => {
     PFEventStrategyFactory.triggerEvent(PFEventsType.SEND_PROFILE);
   }, []);
 
-  const alertButtonStyle: SxProps<Theme> = useIsMobile()
-    ? { textAlign: 'center' }
-    : { textAlign: 'center', minWidth: 'max-content' };
-
-  const handleRedirectToContactsPage = () => {
-    PFEventStrategyFactory.triggerEvent(PFEventsType.SEND_VIEW_CONTACT_DETAILS, {
```
AndreaCimini90, line 41: check that this is event is logged with the new banner logic

AndreaCimini90 (changes requested):

packages/pn-personafisica-webapp/src/pages/Profile.page.tsx
SarahDonvito: Yes, it's logged in Domicile Banner

AndreaCimini90 (approved):

PR #1351, opened 2024-10-09
Fixed wrong copy

leleOFA:
## Short description
Insert correct copy when timeline status is analog failure workflow

in notification utilities file delete control on generetadAarUrl

## How to test
login with PF, username: leonardo in dev enviroment and go on notification with iun : 
YDWX-JQGA-XVTL-202410-H-1

and visualize in timeline correct copy

### DIFF
--- a/packages/pn-commons/src/utility/notification.utility.ts
+++ b/packages/pn-commons/src/utility/notification.utility.ts
@@ -6,7 +6,6 @@
 import _ from 'lodash';
 
 import {
-  AnalogWorkflowDetails,
   ExtRegistriesPaymentDetails,
   F24PaymentDetails,
   INotificationDetailTimeline,
@@ -309,6 +308,12 @@ export function getLegalFactLabel(
   legalFactType?: LegalFactType | 'AAR',
   legalFactKey?: string
 ): string {
+  console.log(
+    'timelineStep,legalFactType,legalFactKey :>> ',
+    timelineStep,
+    legalFactType,
+    legalFactKey
+  );
   const legalFactLabel = getLocalizedOrDefaultLabel(
     'notifications',
     `detail.legalfact`,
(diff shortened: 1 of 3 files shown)


packages/pn-commons/src/utility/notification.utility.ts
```diff
@@ -309,6 +308,12 @@ export function getLegalFactLabel(
   legalFactType?: LegalFactType | 'AAR',
   legalFactKey?: string
 ): string {
+  console.log(
```
AndreaCimini90, line 311: remove this

packages/pn-commons/src/utility/notification.utility.ts
```diff
@@ -386,10 +391,7 @@ export function getLegalFactLabel(
       'detail.timeline.legalfact.analog-failure-delivery',
       'Deposito di avvenuta ricezione'
     );
-  } else if (
-    timelineStep.category === TimelineCategory.ANALOG_FAILURE_WORKFLOW &&
-    (timelineStep.details as AnalogWorkflowDetails).getGeneratedAarUrl
-  ) {
+  } else if (timelineStep.category === TimelineCategory.ANALOG_FAILURE_WORKFLOW) {
```
AndreaCimini90, line 394: Why do you remove the (timelineStep.details as AnalogWorkflowDetails).getGeneratedAarUrl condition?

The TimelineCategory.ANALOG_FAILURE_WORKFLOW can occurs in several ways. Removing the check on generatedAarUrl you are telling that ALL the ANALOG_FAILURE_WORKFLOW must have the same legal fact label and this is wrong!
Only when we have generatedAarUrl, we have an AAR and we have to set this label, in the other cases the default "Attestazione opponibile a terzi" is good

AndreaCimini90 (changes requested):

packages/pn-commons/src/utility/notification.utility.ts
leleOFA: in our case the variable getAarUrl is never evaluated and therefore never enters the if

AndreaCimini90: @leleOFA do you have checked the timeline element that corresponds to the ANALOG_FAILURE_WORKFLOW category?

leleOFA: corresponds to that, in fact by removing the condition that I eliminated in the commit the right copy is inserted

leleOFA: if I insert the condition without the generate but leaving the old one and adding a new one it inserts the copy correctly

AndreaCimini90 (approved):

PR #1384, opened 2024-10-31
fix QR login error - PN-12155

leleOFA:
## Short description
fix error toast when go on notification with aar login from QR
- insert in rejectWithValue null

## How to test
https://cittadini.uat.notifichedigitali.it/auth/login?aar=VEFXUC1SSFpELUVIVEUtMjAyNDA3LUctMV9QRi1iMzJlNDkyMC02ZmYzLTQ4NzItODAxOC1kNjBhNGU1ODI3ZjlfYTY2Yjk0NWEtMDM5OS00Mzg1LWE0N2EtZjU3MzhlNjg2MzVm

insert wrong credentials when login


packages/pn-personafisica-webapp/src/redux/notification/actions.ts
```diff
@@ -275,7 +275,7 @@ export const exchangeNotificationQrCode = createAsyncThunk<BffCheckAarResponse,
       const response = await notificationReceivedApiFactory.checkAarQrCodeV1(params);
       return response.data;
     } catch (e: any) {
-      return rejectWithValue(parseError(e));
```
mflauti, line 278: Restore the old version and use the AppResponsePublisher to intercept the error and manage it properly. Do the same for PG.

mflauti (changes requested):

packages/pn-personafisica-webapp/src/redux/notification/actions.ts
AndreaCimini90: AppResponsePublisher can be useful when we have to manage the error in a way that is different from the standard way (i.e. show the toast error).
In this case, it's ok to show the error but we have to show it with a different message. So ask to Daniela or Norma the copy (tag them in the issue) and after manage the error through the AppError

AndreaCimini90 (changes requested):

packages/pn-personafisica-webapp/src/navigation/AARGuard.tsx
```diff
@@ -52,6 +52,14 @@ const AARGuard = () => {
     }
   }, [aar]);
 
+  useEffect(() => {
+    AppResponsePublisher.error.subscribe('exchangeNotificationQrCode', ()=>false);
```
AndreaCimini90, line 56: the code is ok. we have to manage only the case of 404 and PN_DELIVERY_NOTIFICATIONNOTFOUND. In the function ()=>false, you should have the error object and you can use it to check if the error is the error with code PN_DELIVERY_NOTIFICATIONNOTFOUND

AndreaCimini90 (changes requested):

packages/pn-personafisica-webapp/src/navigation/AARGuard.tsx
leleOFA: doine

AndreaCimini90 (approved):

PR #1392, opened 2024-11-05
Refactor/PN-13040: refactor 3rd step "attachments" in notification creation

fbianchicodermine:
## Short description
Refactor of UI third tab of manual notification according to the mockup.

- Hash code in `FileUpload` component with copy button

## How to test
Start pa-webapp. Create manual notification and check 3rd tab


packages/pn-pa-webapp/src/components/NewNotification/Attachments.tsx
```diff
@@ -49,14 +56,18 @@ const AttachmentBox: React.FC<AttachmentBoxProps> = ({
...
-  fieldErros,
+  fieldErrors,
   onFieldTouched,
   onFileUploaded,
   onRemoveFile,
   fileUploaded,
 }) => {
+  const [focused, setFocused] = useState(false);
   const { t } = useTranslation(['notifiche']);
+  const { t: tc } = useTranslation(['common']);
```
AndreaCimini90, line 67: Avoid to define another t function. Inlcude common in the array above and use the ns prop

packages/pn-pa-webapp/src/components/NewNotification/Attachments.tsx
```diff
@@ -49,14 +56,18 @@ const AttachmentBox: React.FC<AttachmentBoxProps> = ({
...
   fieldValue,
   fieldTouched,
-  fieldErros,
+  fieldErrors,
   onFieldTouched,
   onFileUploaded,
   onRemoveFile,
   fileUploaded,
 }) => {
+  const [focused, setFocused] = useState(false);
```
AndreaCimini90, line 65: Why we need this?

AndreaCimini90 (changes requested):

packages/pn-pa-webapp/src/components/NewNotification/Attachments.tsx
fbianchicodermine: We need to show helper text of max characters always when input is focused

packages/pn-commons/src/components/FileUpload.tsx
```diff
@@ -140,6 +139,12 @@ const FilenameBox = ({ filename }: { filename: string }) => {
   );
 };
 
+const TextFieldDisabled = styled(TextField)(({ theme }) => ({
```
AndreaCimini90, line 142: Why we need this? Generally we avoid to customize mui component

packages/pn-pa-webapp/src/components/NewNotification/Attachments.tsx
AndreaCimini90: Ok we can avoid the use of useState to prevent unnecessary re-rendering. Try to use the [useFormControl](https://mui.com/material-ui/react-text-field/#useformcontrol)

AndreaCimini90 (changes requested):

packages/pn-commons/src/components/FileUpload.tsx
fbianchicodermine: I want to change color of disabled TextField. It's not an editable field but in mockup it's designed with style as an input.

AndreaCimini90: Ok I understood. We can use readonly property instead of disabled, and avoid customization. What do you think?

fbianchicodermine: Yes, great!

AndreaCimini90 (approved):

PR #1394, opened 2024-11-06
Test/PN-12898: test new form components

fbianchicodermine:
## Short description
Create specific test for new components PreliminaryInformationsLang e PreliminaryInformationsContent

## How to test
Run test

### DIFF
--- a/packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsContent.test.tsx
+++ b/packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsContent.test.tsx
@@ -1,0 +1,131 @@
+import { Formik, FormikProps } from 'formik';
+import { vi } from 'vitest';
+
+import {
+  fireEvent,
+  getById,
+  render,
+  screen,
+  testFormElements,
+} from '@pagopa-pn/pn-commons/src/test-utils';
+
+import { NewNotificationLangOther } from '../../../models/NewNotification';
+import { PreliminaryInformationsPayload } from '../../../redux/newNotification/types';
+import PreliminaryInformationsContent from '../PreliminaryInformationsContent';
+
+// mock imports
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (str: string) => str,
+    i18n: { language: 'it' },
+  }),
+}));
+
+describe('PreliminaryInformationsContent', () => {
+  const mockFormikProps = {
+    values: {
+      lang: 'it',
+      subject: '',
+      abstract: '',
+      additionalLang: '',
+      additionalSubject: '',
+      additionalAbstract: '',
+    },
+    touched: {},
+    errors: {},
+    handleChange: jest.fn(),
+  } as unknown as FormikProps<PreliminaryInformationsPayload>;
+
+  const mockLanguages = {
+    it: 'Italiano',
+    de: 'Tedesco',
+  };
+
+  const mockOnChangeTouched = jest.fn();
+  const mockSubjectHelperText = 'Helper text for subject';
+
+  it('renders the component', () => {
+    const result = render(
+      <Formik initialValues={mockFormikProps.values} onSubmit={() => {}}>
+        <PreliminaryInformationsContent
+          formik={mockFormikProps}
+          languages={mockLanguages}
+          onChangeTouched={mockOnChangeTouched}
+          subjectHelperText={mockSubjectHelperText}
+        />
+      </Formik>
+    );
+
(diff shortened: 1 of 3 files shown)


packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsContent.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const mockSubjectHelperText = 'Helper text for subject';
+
+  it('renders the component', () => {
+    const result = render(
+      <Formik initialValues={mockFormikProps.values} onSubmit={() => {}}>
+        <PreliminaryInformationsContent
+          formik={mockFormikProps}
+          languages={mockLanguages}
+          onChangeTouched={mockOnChangeTouched}
+          subjectHelperText={mockSubjectHelperText}
```
AndreaCimini90, line 54: Why are we passing the subjectHelperText when we pass the formik instance to the component?

packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsContent.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+      subject: '',
+      abstract: '',
+      additionalLang: '',
+      additionalSubject: '',
+      additionalAbstract: '',
+    },
+    touched: {},
+    errors: {},
+    handleChange: jest.fn(),
+  } as unknown as FormikProps<PreliminaryInformationsPayload>;
```
AndreaCimini90, line 37: We mock data in rare case, because mocking can lead to a cases that are different from the real ones. So instead of mocking, create a test component here and in it init formik and pass it to the PreliminaryInformationsContent

packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsLang.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+}));
+
+const mockFormikProps = {
+  values: {
+    lang: 'it',
+    additionalLang: '',
+  },
+  handleChange: jest.fn(),
+  setFieldTouched: jest.fn(),
+} as unknown as FormikProps<PreliminaryInformationsPayload>;
```
AndreaCimini90, line 36: Same as before

packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsContent.test.tsx
fbianchicodermine: I always need to show helper text when input is focused, not only when there’s an error

AndreaCimini90: and the formik istance is not enough?

packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsContent.test.tsx
fbianchicodermine: Ok, I'll keep it in mind for the next tests

packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsContent.test.tsx
fbianchicodermine: I don't have that string in formik

packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsContent.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+            subjectHelperText={mockSubjectHelperText}
+          />
+        )}
+      </Formik>
+    );
+
+    const subjectInput = getById(container, 'subject');
+    fireEvent.focus(subjectInput);
+
+    expect(getByText(container, mockSubjectHelperText)).toBeInTheDocument();
```
AndreaCimini90, line 99: Don't import getByText but use that from the return of render function. Do it in all tests

packages/pn-pa-webapp/src/components/NewNotification/__test__/PreliminaryInformationsContent.test.tsx
AndreaCimini90: Ok and we can't move that string inside the component?

AndreaCimini90 (approved):

PR #1439, opened 2025-01-13
Fix/PN-13596: [a11y] notification status tooltip on mobile

fbianchicodermine:
## Short description
This PR resolves problem of badge status of notification for screen reader on mobile

- Remove `clickawaylistener` and a `box` in custom tooltip
- Add `aria-label` that describes the info of status chip (`<chip text>: <tooltip text>`)

## How to test
Go to page notification list in mobile mode. Navigate page with tab and activate screen reader. Check that screen reader reads chiop and content of tooltip


packages/pn-commons/src/components/Notifications/StatusTooltip.tsx
```diff
@@ -24,10 +24,14 @@ const StatusTooltip = ({
     <CustomTooltip openOnClick={false} tooltipContent={tooltipContent} tooltipProps={tooltipProps}>
       <Chip
         id={`status-chip-${label}`}
-        label={label}
+        label={<span aria-hidden="true">{label}</span>}
```
SarahDonvito, line 27: @fbianchicodermine use conditional label: if mobile use span instead use only label

SarahDonvito (changes requested):

packages/pn-commons/src/components/Notifications/StatusTooltip.tsx
fbianchicodermine: This behavior also applies to desktop. Otherwise, VoiceOver does not read the tooltip content.

SarahDonvito (approved):

PR #1477, opened 2025-02-19
Fix/PN-13388: improve regex for redirect and robots header

fbianchicodermine:
## Short description
Include a summary of the changes.

## How to test
Start app (ex: cittadini) and check in html header "meta robots" has "noindex"

### DIFF
--- a/packages/pn-pa-webapp/public/js/config.js
+++ b/packages/pn-pa-webapp/public/js/config.js
@@ -1,4 +1,4 @@
-const regex = new RegExp('https://selfcare(.(dev|test|uat|hotfix))?.notifichedigitali.it');
+const regex = new RegExp('https://selfcare\.(((dev|test|uat|hotfix)\.)?)notifichedigitali\.it') 
 const origin = window.origin;
 if (regex.test(origin)) {
   document.write(`<meta name="robots" content="noindex">`);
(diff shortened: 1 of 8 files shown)


packages/pn-pa-webapp/public/js/config.js
```diff
@@ -1,4 +1,4 @@
-const regex = new RegExp('https://selfcare.((dev|test|uat|hotfix).?)notifichedigitali.it');
+const regex = new RegExp('https://selfcare(.(dev|test|uat|hotfix))?.notifichedigitali.it');
```
AndreaCimini90, line 1: I think that the problem is the use of char "." without escaping it. This char is a special char in regex and if we want that it is treated as a point we have to escape it with back slash.

So we can restore the old regex and escape the char

AndreaCimini90 (changes requested):

packages/pn-pa-webapp/public/js/config.js
fbianchicodermine: The old regex is wrong. It allows strings like `https://cittadini.devnotifichedigitali.it`. 
It's not a problem with char "."

fbianchicodermine: Fix it with "\" before "."

AndreaCimini90 (approved):

AndreaCimini90 (approved):

PR #1497, opened 2025-03-06
PN-14148 - Update formik recipient key to handle numeric CF

alessandrogelmi:
## Short description
Update formik recipient key to handle numeric CF by adding `recipientType` before `taxId`

- Edit formik `initialValues` and `validationSchema`

## How to test
- Start PA
- Create a new notification with only a PG as recipient (or a PF and a PG)
- Select PagoPA as debt position for both recipients
- Try to insert the same `noticeCode` and check that error is show. Same for the applyCost switch

### DIFF
--- a/packages/pn-pa-webapp/src/components/NewNotification/DebtPositionDetail.tsx
+++ b/packages/pn-pa-webapp/src/components/NewNotification/DebtPositionDetail.tsx
@@ -1,5 +1,5 @@
 import { useFormik } from 'formik';
-import _, { mapValues } from 'lodash';
+import _ from 'lodash';
 import { ChangeEvent, ForwardedRef, forwardRef, useImperativeHandle, useMemo } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import * as yup from 'yup';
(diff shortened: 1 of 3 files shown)


packages/pn-pa-webapp/src/components/NewNotification/PaymentMethods.tsx
```diff
@@ -126,9 +126,10 @@ const PaymentMethods: React.FC<Props> = ({
         if (recipient.debtPosition === PaymentModel.NOTHING) {
           return <></>;
         }
+        const recipientIndex = `${recipient.recipientType}-${recipient.taxId}`;
```
fbianchicodermine, line 129: The name `recipientIndex` is ambiguous. It's not an index.

packages/pn-pa-webapp/src/components/NewNotification/DebtPositionDetail.tsx
```diff
@@ -109,7 +109,8 @@ const DebtPositionDetail: React.FC<Props> = ({
   const formatPayments = (): Array<NewNotificationRecipient> => {
     const recipients = _.cloneDeep(notification.recipients);
     return recipients.map((recipient) => {
-      const recipientData = formik.values.recipients[recipient.taxId];
+      const recipientIndex = `${recipient.recipientType}-${recipient.taxId}`;
```
fbianchicodermine, line 112: The name `recipientIndex` is ambiguous. It's not an index.

fbianchicodermine (changes requested):

packages/pn-pa-webapp/src/components/NewNotification/PaymentMethods.tsx
alessandrogelmi: Why? It represents the index of the recipient inside the `recipients` map.

packages/pn-pa-webapp/src/components/NewNotification/DebtPositionDetail.tsx
alessandrogelmi: same as above

fbianchicodermine (approved):

PR #1506, opened 2025-03-10
fix(PN-14145): fix copy as figma

SarahDonvito:
## Short description
fix copy as figma

## How to test
Login as PA, check in create notification > payment section is present the type of file to upload json or pdf

### DIFF
--- a/packages/pn-pa-webapp/public/locales/it/notifiche.json
+++ b/packages/pn-pa-webapp/public/locales/it/notifiche.json
@@ -401,7 +401,7 @@
     },
     "back-to-notifications": "Torna alle notifiche",
     "drag-doc-pc": "Trascina qui il documento",
-    "drag-payment-doc-pc": "Trascina qui il documento in formato {{type}}",
+    "drag-doc-with-format-pc": "Trascina qui il documento in formato {{type}}",
     "drag-doc-mobile": "Carica qui il documento",
     "warning-payment-disabled": "Al momento le notifiche che prevedono un pagamento si possono inviare solo tramite API Key.",
     "steps": {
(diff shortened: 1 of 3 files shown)


alessandrogelmi (approved):

packages/pn-pa-webapp/src/components/NewNotification/F24PaymentBox.tsx
```diff
@@ -58,7 +58,9 @@ const F24PaymentBox: React.FC<PaymentBoxProps> = ({
       <FileUpload
         data-testid="fileUploadInput"
         uploadText={
-          isMobile ? t('new-notification.drag-doc-mobile') : t('new-notification.drag-doc-pc')
+          isMobile
+            ? t('new-notification.drag-doc-mobile')
+            : t('new-notification.drag-payment-doc-pc', { type: '.json' })
```
alessandrogelmi, line 63: On Figma `json` is uppercase

packages/pn-pa-webapp/public/locales/it/notifiche.json
```diff
@@ -401,6 +401,7 @@
     },
     "back-to-notifications": "Torna alle notifiche",
     "drag-doc-pc": "Trascina qui il documento",
+    "drag-payment-doc-pc": "Trascina qui il documento in formato {{type}}",
```
alessandrogelmi, line 404: I should rename it with `drag-doc-with-format-pc`

alessandrogelmi (changes requested):

packages/pn-pa-webapp/src/components/NewNotification/F24PaymentBox.tsx
SarahDonvito: I think we should use lowercase since is the extension of the file

alessandrogelmi (approved):

PR #1507, opened 2025-03-11
fix warning with formhelpertextprops

fbianchicodermine:


packages/pn-personafisica-webapp/src/components/Notifications/FilterNotificationsFormBody.tsx
```diff
@@ -91,11 +91,10 @@ const FilterNotificationsFormBody = ({
           helperText={
             formikInstance.touched.iunMatch &&
             formikInstance.errors.iunMatch && (
-              <FormHelperText error aria-live="assertive">
-                {String(formikInstance.errors.iunMatch)}
-              </FormHelperText>
+                String(formikInstance.errors.iunMatch)
```
AndreaCimini90, line 94: Should be useless

packages/pn-personagiuridica-webapp/src/components/Notifications/FilterNotificationsFormBody.tsx
```diff
@@ -90,12 +90,10 @@ const FilterNotificationsFormBody = ({
           error={formikInstance.touched.iunMatch && Boolean(formikInstance.errors.iunMatch)}
           helperText={
             formikInstance.touched.iunMatch &&
-            formikInstance.errors.iunMatch && (
-              <FormHelperText error aria-live="assertive">
-                {String(formikInstance.errors.iunMatch)}
-              </FormHelperText>
-            )
+            formikInstance.errors.iunMatch &&
+            String(formikInstance.errors.iunMatch)
```
AndreaCimini90, line 94: same as before

AndreaCimini90 (changes requested):

packages/pn-personafisica-webapp/src/components/Notifications/FilterNotificationsFormBody.tsx
fbianchicodermine: It is used for TS typing. `helperText` wants a string

AndreaCimini90: I meant that we can remove String(formikInstance.errors.iunMatch)

PR #1509, opened 2025-03-14
fix/PN-14240 - Digital Domicile Bug Fix

mflauti:
## Short description
This PR fixes some bugs reported on Jira tasks PN-14240, PN-14241 and PN-14218

- fix a bug which was causing to advance to the following step instead of skipping any remaining step when clicking on the exit after choosing sercq as default legal contact during ddom activation/transfer wizard
- fix a bug not showing any error text typing in a wrong or partial sender name during a special contact add. Also set the proper sender id on input change
- fix cosmetic bug setting the proper margin and a Divider element to EmailSmsContactWizard component

## How to test
Run PF and PG web applications and check every bug is resolved as indicated inside the related Jira task


packages/pn-personafisica-webapp/src/components/Contacts/DigitalContactActivation.tsx
```diff
@@ -58,30 +59,38 @@ const DigitalContactActivation: React.FC<Props> = ({ isTransferring = false, onG
...
-      showConfirmationModal();
+      showConfirmationModal(exit);
     }
   };
 
-  const showConfirmationModal = () => {
+  const showConfirmationModal = (exit?: boolean) => {
     if (activeStep === 1 && showIOStep) {
-      setModal({ open: true, step: ActiveStep.IO });
+      setModal({ open: true, step: ActiveStep.IO, exit });
```
fbianchicodermine, line 91: we can use `const step = ...` and not repeat function setModal

fbianchicodermine (changes requested):

fbianchicodermine (approved):

PR #1510, opened 2025-03-18
fix/PN-13660 - Hide 'name' column on VirtualKeyTable for non-admin users (PG)

mflauti:
## Short description
Hide 'name' field for personal keys when the user is entitled to only view his/her own keys (non-admin)

## How to test
On PG webapp verify that only admin can see the 'name' column inside the API Integration section but admin with groups and operator cannot

### DIFF
--- a/packages/pn-personagiuridica-webapp/src/components/IntegrazioneApi/VirtualKeysTable.tsx
+++ b/packages/pn-personagiuridica-webapp/src/components/IntegrazioneApi/VirtualKeysTable.tsx
@@ -108,6 +108,8 @@ const VirtualKeysTable: React.FC<Props> = ({
     },
   ];
 
+  const VirtualKeysVisibleColumns = isUserAdmin ? virtualKeysColumns : virtualKeysColumns.slice(1);
+
   if (!virtualKeys || virtualKeys.items.length === 0) {
     return (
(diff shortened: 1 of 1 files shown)


packages/pn-personagiuridica-webapp/src/components/IntegrazioneApi/VirtualKeysTable.tsx
```diff
@@ -108,6 +108,8 @@ const VirtualKeysTable: React.FC<Props> = ({
     },
   ];
 
+  const VirtualKeysVisibleColumns = isUserAdmin ? virtualKeysColumns : virtualKeysColumns.slice(1);
```
alessandrogelmi, line 111: The initial letter of const variable should be lowercase

alessandrogelmi (changes requested):

packages/pn-personagiuridica-webapp/src/components/IntegrazioneApi/VirtualKeysTable.tsx
mflauti: Absolutely! I'll fix it immediately!

alessandrogelmi (approved):

alessandrogelmi (approved):

PR #1512, opened 2025-03-19
fix(PN-14217) Errore non conservato con cambio pagina, creazione notifica

leleOFA:
## Short description
insert fix for form

- insert control of value if also exist


packages/pn-pa-webapp/src/components/NewNotification/DebtPositionDetail.tsx
```diff
@@ -77,6 +77,12 @@ const DebtPositionDetail: React.FC<Props> = ({
   const { PAYMENT_INFO_LINK } = getConfiguration();
   const dispatch = useAppDispatch();
 
+  const showErrorIfPresent = (fieldName: string) =>
```
fbianchicodermine, line 80: use a better name for this function. Es: `hasFieldError`

fbianchicodermine (changes requested):

PR #1521, opened 2025-03-28
PN-14412 - Add organization logo to header on PG

alessandrogelmi:
## Short description
Add organization logo to header of PG webapp

- Add organization logo to header
- Create new configuration variable `SELFCARE_CDN_URL`


fbianchicodermine (changes requested):
I would implement a test that checks for the existence of an `` with that href.

PR #1608, opened 2025-07-24
fix(semantic-html): avoid <p> inside <p> in filter notifications

fbianchicodermine:
## Short description
specified component "span" for FormHelperText to avoid a <p> inside <p>


packages/pn-personagiuridica-webapp/src/components/Notifications/FilterNotificationsFormBody.tsx
```diff
@@ -130,7 +130,7 @@ const FilterNotificationsFormBody = ({
                 'data-testid': 'input(start date)',
               },
               helperText: (
-                <FormHelperText error aria-live="assertive">
+                <FormHelperText sx={{ ml: 0 }} error aria-live="assertive" component="span">
```
AndreaCimini90, line 133: same as before

AndreaCimini90 (approved):

PR #1615, opened 2025-08-06
fix(PN-16022): include .well-known folder in tar build

fbianchicodermine:
## Short description
With command `tar c *` hidden files are not included. We need to add .well-known folder.


AndreaCimini90 (commented):
Questo è un commento di prova

AndreaCimini90 (approved):

PR #1808, opened 2026-04-02
fix(A2-2352): add semantic input attributes to contact fields

mflauti:


packages/pn-personafisica-webapp/src/components/Contacts/DigitalContact.tsx
```diff
@@ -128,6 +128,26 @@ const DigitalContact = forwardRef<{ toggleEdit: () => void }, Props>(
       },
     });
 
+    // Semantic/native input attributes derived from channel type.
+    const getSemanticTextFieldProps = (): Partial<TextFieldProps> => {
```
AndreaCimini90: To avoid useless rerendering, can we put this function outside the compnent inside the contacts.utility file?

packages/pn-personagiuridica-webapp/src/components/Contacts/DigitalContact.tsx
```diff
@@ -120,6 +120,26 @@ const DigitalContact = forwardRef<{ toggleEdit: () => void }, Props>(
       },
     });
 
+    // Semantic/native input attributes derived from channel type.
+    const getSemanticTextFieldProps = (): Partial<TextFieldProps> => {
```
AndreaCimini90: Same as before

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1816, opened 2026-04-09
fix(PN-18026): save rapid access params in local storage on iOS devices

alessandrogelmi:
## Short description

Persist rapid access parameters (`aar` and `retrievalId`) in local storage instead of session storage when the user is on an iOS mobile device.

- Store rapid access parameters in local storage for iOS mobile users
- Clear local storage on login success or error, on logout, and on a new login attempt to prevent stale data
- Fix unit tests

## How to test

- Start PF login and PF web app locally  
- Verify that the current SPID flow still works correctly  
- Attempt a login flow with `aar` or `retrievalId` in the query parameters and force the login to fail. Then click "Go to login" and verify that you are redirected back to the login page with `aar` or `retrievalId` still present in the query parameters


packages/pn-personafisica-login/src/utility/storage.ts
```diff
@@ -1,11 +1,13 @@
 import { AppRouteParams, storageOpsBuilder } from '@pagopa-pn/pn-commons';
 
-const storageAarOps = storageOpsBuilder<string>(AppRouteParams.AAR, 'string', false);
+import { isIOSMobile } from './utils';
+
+const storageAarOps = storageOpsBuilder<string>(AppRouteParams.AAR, 'string', isIOSMobile());
```
AndreaCimini90: We have to limit the localStorage use to only case in witch we are on IOS and the user select CIE. So we have to move the storage write to the idp selection

packages/pn-personafisica-login/src/utility/utils.tsx
```diff
@@ -11,3 +11,5 @@ export const shuffleList = (list: Array<any>) => {
 };
 
 export const generateRandomUniqueString = () => uuidv4().replace(/-/g, '').slice(0, 20);
+
+export const isIOSMobile = (): boolean => /iPhone|iPod|iPad/.test(navigator.userAgent);
```
AndreaCimini90: In this article https://dev.to/konyu/using-javascript-to-determine-whether-the-client-is-ios-or-android-4i1j the check on userAgent isn't enough. Can you go into detail?

packages/pn-personafisica-login/src/pages/loginError/LoginError.tsx
```diff
@@ -8,8 +8,10 @@ import { IllusError } from '@pagopa/mui-italia';
 import { PFLoginEventsType } from '../../models/PFLoginEventsType';
 import { ROUTE_LOGIN } from '../../navigation/routes.const';
 import PFLoginEventStrategyFactory from '../../utility/MixpanelUtils/PFLoginEventStrategyFactory';
+import { storageRapidAccessOps } from '../../utility/storage';
 
 const handleError = (queryParams: string, errorMessage: string) => {
+  storageRapidAccessOps.delete();
```
AndreaCimini90, line 15: It's good to remove the rapid access stored, but I should store it here and I should append to the login page if the user clicks the button "Go to login"

AndreaCimini90 (changes requested):

packages/pn-personafisica-login/src/utility/storage.ts
```diff
@@ -1,37 +1,39 @@
...
-const storageRetrievalIdOps = storageOpsBuilder<string>(
-  AppRouteParams.RETRIEVAL_ID,
-  'string',
-  false
-);
+const storageRetrievalIdOps = (useLocalStorage: boolean) =>
+  storageOpsBuilder<string>(AppRouteParams.RETRIEVAL_ID, 'string', useLocalStorage);
 
 export const storageRapidAccessOps = {
   read: (): [AppRouteParams, string] | undefined => {
```
AndreaCimini90, line 10: I think it is better that also read and delete have the useLocalStorage param

packages/pn-personafisica-login/src/utility/storage.ts
```diff
@@ -1,37 +1,39 @@
...
-    const retrievalId = storageRetrievalIdOps.read();
+    const retrievalId = storageRetrievalIdOps(false).read() ?? storageRetrievalIdOps(true).read();
     if (retrievalId) {
       return [AppRouteParams.RETRIEVAL_ID, retrievalId];
     }
     return undefined;
   },
-  write: ([key, value]: [AppRouteParams, string]) => {
+  write: ([key, value]: [AppRouteParams, string], useLocalStorage = false) => {
+    storageAarOps(false).delete();
```
AndreaCimini90, line 22: here I should restore the old version and, instead of removing both session and local storage, I should use useLocalStorage param

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1832, opened 2026-04-21
feat(PN-19473): Add onboarding reminder modal on notification detail exit

mflauti:
## Short description
Add onboarding exit reminder modal in notification detail when user leaves without required contacts or after returning from payment.

- add onboarding exit reminder modal in NotificationDetail
- show modal on navigation attempt when onboarding conditions are met
- show modal on return from payment when onboarding conditions are met
- prevent modal close on backdrop click (allow close via buttons/ESC only)

## How to test
- open a notification detail without required contacts configured
- try to navigate back to the notifications list -> modal should appear
- simulate a return from payment (or complete a payment flow) -> modal should appera
- verify that:
  - clicking "Configure" navigates to onboarding
  - clicking "Skip" closes the modal (after payment) or continues to notifications list
  - clicking outside the modal does nothing
  - the modal is not shown again in the same session after being skipped or after redirecting to onboarding
  - the modal is not shown when the feature flag is disabled


packages/pn-commons/src/components/ConfirmationModal.tsx
```diff
@@ -53,7 +53,13 @@ const ConfirmationModal: React.FC<Props> = ({
   return (
     <PnDialog
       open={open}
-      onClose={slotsProps?.closeButton?.onClick}
+      onClose={(_, reason) => {
+        if (reason === 'backdropClick') {
```
AndreaCimini90: Why do we need this?

packages/pn-personafisica-webapp/src/redux/onboarding/reducers.ts
```diff
@@ -0,0 +1,56 @@
+import { PayloadAction, createSelector, createSlice } from '@reduxjs/toolkit';
```
AndreaCimini90: this was moved to the sidemenu reducer

packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -120,6 +126,75 @@ const NotificationDetail: React.FC = () => {
   const userPayments = useAppSelector((state: RootState) => state.notificationState.paymentsData);
   const paymentTpp = useAppSelector((state: RootState) => state.generalInfoState.paymentTpp);
 
+  // START ONBOARDING
+  const addresses = useAppSelector(contactsSelectors.selectAddresses);
```
AndreaCimini90: Pay attention to show the modal when the user is mandate. In that case we don't need to show the modal because also even if the mandate sets his contacs, he will never be informed of new notifications that are not his. So inviting him to add contacts for notifications that aren't his makes no sense.

packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -750,6 +825,31 @@ const NotificationDetail: React.FC = () => {
               </Box>
             </Grid>
           </Grid>
+          <ConfirmationModal
```
AndreaCimini90: Is it possible to move al the onboarding blocker logic into a component as done for Prompt component?

AndreaCimini90 (changes requested):

packages/pn-commons/src/components/ConfirmationModal.tsx
mflauti: Because otherwise backdrop clicks would behave like the "Skip"/close action. This avoids accidental dismissals and requires an explicit user choice, while still allowing ESC to close the modal.

packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
mflauti: Good point. I did not find this explicitly stated in the SRS (we could consider updating it), but it makes sense.
I will update the logic to avoid showing the modal in this case.

packages/pn-commons/src/components/ConfirmationModal.tsx
mflauti: We decided to move this logic from the shared `ConfirmationModal` to `NotificationDetailOnboardingPrompt`.

AndreaCimini90 (approved):

PR #1839, opened 2026-04-22
fix(PN-19275): solved some issues and avoid to call notifications api at first start 

AndreaCimini90:
## Short description
Fixed various bug

- Reduced images size
- Reset general state on logout
- Fixed modal layout
- Hide IO package when there is at least one courtesy address
- Avoid redirect to onboarding page if feature flag is false
- Avoid double call to notifications list api

## How to test
Login with PF and test all the bug fixed


packages/pn-personafisica-webapp/src/App.tsx
```diff
@@ -282,6 +285,7 @@ const App = () => {
       void dispatch(getSidemenuInformation());
       void dispatch(getCurrentAppStatus());
       void dispatch(getReceivedNotifications({ size: 10 })); // to do: remove the first one from notifications page
```
fbianchicodermine: What does the comment mean? If it's old, remove it

fbianchicodermine (changes requested):

packages/pn-personafisica-webapp/src/App.tsx
AndreaCimini90: done

fbianchicodermine (approved):

PR #1859, opened 2026-04-29
fix(PN-15937): move max-attachments label into AttachmentBox and fix …

gaetano-amato:
## Short description
Moved the "max-attachments" label inside the AttachmentBox component and fixed its layout to ensure correct vertical alignment with the rest of the form elements.

- moved "max-attachments" label from parent component to AttachmentBox
- added a new prop to AttachmentBox to handle conditional rendering of the label
- displayed the label only for the first attachment
- updated layout to enforce column alignment of elements
- removed duplicate label from parent component

## How to test

- Open the "Invia notifica" flow
- Navigate to the attachments step
- Verify that:
- the "max-attachments" label is displayed under the first attachment box
- the label is not duplicated for other attachments


alessandrogelmi (changes requested):
Fix the `useMemo` inside `NewNotification.page.tsx` and the copy about attachments in the last step

alessandrogelmi (approved):

PR #1887, opened 2026-05-14
test: replace Object.defineProperty with vi.stubGlobal

fbianchicodermine:
## Short description
- Replace Object.defineProperty(globalThis, ...) with vi.stubGlobal(...) across all test files in pn-personafisica-webapp, pn-personagiuridica-webapp, pn-pa-webapp, pn-commons, and pn-personafisica-login
- Remove async describe anti-pattern from all affected test files

## Motivation
Object.defineProperty on top-level globals (location, open) does not restore the original descriptor after the test, causing state to leak between test files and making tests order-dependent. vi.stubGlobal is the Vitest-idiomatic alternative: it snapshots the original value and restores it automatically via vi.unstubAllGlobals() in afterAll, regardless of pool type.

async describe is an anti-pattern in Vitest: the test runner expects describe callbacks to be synchronous. An async callback causes beforeAll/afterAll lifecycle hooks to be registered after the runner has already moved on, leading to deadlocks or skipped teardown — particularly visible with vmThreads pool.

## How to test
Run full test suite across all affected packages and verify no regressions

### DIFF
--- a/packages/pn-pa-webapp/src/pages/__test__/AppStatus.page.test.tsx
+++ b/packages/pn-pa-webapp/src/pages/__test__/AppStatus.page.test.tsx
@@ -25,17 +25,19 @@ describe('AppStatus page', () => {
 
   beforeAll(() => {
     mock = new MockAdapter(apiClient);
+  });
+
+  beforeEach(() => {
     vi.stubGlobal('location', { href: '' });
   });
 
   afterEach(() => {
     mock.reset();
-    globalThis.location.href = '';
+    vi.unstubAllGlobals();
   });
 
   afterAll(() => {
     mock.restore();
-    vi.unstubAllGlobals();
   });
   /*
    * The intent of the "OK" test is to verify somehow that the result of the API calls
--- a/packages/pn-pa-webapp/src/pages/__test__/AppStatus.page.test.tsx
+++ b/packages/pn-pa-webapp/src/pages/__test__/AppStatus.page.test.tsx
@@ -1,4 +1,5 @@
 import MockAdapter from 'axios-mock-adapter';
+import { vi } from 'vitest';
 
 import { ThemeProvider } from '@mui/material';
 import { AppResponseMessage, ResponseEventDispatcher, formatDate } from '@pagopa-pn/pn-commons';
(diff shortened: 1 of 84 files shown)


packages/pn-pa-webapp/src/pages/__test__/AppStatus.page.test.tsx
```diff
@@ -19,26 +20,22 @@ const AppStatusWithErrorHandling = () => (
...
-      configurable: true,
-      value: { href: '' },
-    });
+    vi.stubGlobal('location', { href: '' });
   });
 
   afterEach(() => {
     mock.reset();
-    window.location.href = '';
+    globalThis.location.href = '';
```
AndreaCimini90: We don't have a vi method to do this?

alessandrogelmi (dismissed):

packages/pn-personafisica-webapp/src/components/Contacts/__test__/IOContact.test.tsx
```diff
@@ -130,25 +126,25 @@ describe('IOContact component', async () => {
...
-    Object.defineProperty(globalThis, 'navigator', {
-      value: { userAgent: 'iPhone' },
-    });
+    vi.stubGlobal('navigator', { userAgent: 'iPhone' });
     const { getByRole } = render(<IOContact />);
     const button = getByRole('button', { name: 'io-contact.download' });
     fireEvent.click(button);
     expect(assignFn).toHaveBeenCalledTimes(1);
     expect(assignFn).toHaveBeenCalledWith(getConfiguration().APP_IO_IOS);
+    vi.unstubAllGlobals();
```
AndreaCimini90: we cannot do the unstub only for a specific mock?

AndreaCimini90 (changes requested):

packages/pn-pa-webapp/src/pages/__test__/AppStatus.page.test.tsx
fbianchicodermine: No. 
We can put `vi.stubGlobal` in beforeEach and `vi.unstubAllGlobals` in afterEach.

packages/pn-personafisica-webapp/src/components/Contacts/__test__/IOContact.test.tsx
fbianchicodermine: Vitest doesn't expose a vi.unstubGlobal(key) API. The only built-in option is vi.unstubAllGlobals()

AndreaCimini90: Ok! We can move stub and unstub into beforeEach and afterEach? So we avoid to repeat then every test case

fbianchicodermine: ok! done

AndreaCimini90 (approved):

PR #1891, opened 2026-05-15
fix(PN-19419): enhance mixpanel tracking with transport option

alessandrogelmi:
## Short description

Add optional tracking options to `mixpanel.track()` functions in order to use `transport: "sendBeacon"` for the SEND_IDP_SELECTED and SEND_LOGIN_METHOD Mixpanel events. This property ensures that the events will be tracked even if the current page is closed or navigated away from.

- Add optional tracking options to the Mixpanel tracking function

## How to test

- Start PF login
- Accept cookies so tracking uses the real Mixpanel APIs
- Select "Entra con CIE" or a SPID entity — the browser will navigate to the IDP page
- Go back to the PF login page and check the console to verify that the event has been sent


packages/pn-commons/src/services/tracking.service.ts
AndreaCimini90: set trackOptions also here

packages/pn-personafisica-login/src/pages/success/Success.tsx
AndreaCimini90: this should be done after the event sending

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1893, opened 2026-05-15
feat(PN-19759): Add Mixpanel functional tracking foundation

mflauti:
## Short description
Add common tracking helpers and a typed functional event registry while preserving compatibility with the existing PF strategy-based tracking.
Implement the first subset of PG events to validate the new approach.

- Add common tracking helpers in commons for standard Mixpanel events
- Make `EventStrategyFactory` compatible with typed payload maps while keeping PF backward compatibility
- Add PF functional tracking registry and strategy adapter
- Implement the first PG tracked events for notifications and mandate creation
- Add focused unit test for common helpers, PG domain mapping, registry composition, strategy adapter

## How to test
Run PG app and:
- go to notifications list → SEND_PG_YOUR_NOTIFICATION should trigger
- open a notification detail → SEND_PG_NOTIFICATION_DETAIL should trigger
- go to mandates / DelegatesByCompany and click "Add a new mandate" → SEND_PG_ADD_MANDATE_START should trigger
- send a new mandate → SEND_PG_ADD_MANDATE_UX_SUCCESS should trigger


packages/pn-personagiuridica-webapp/src/pages/NuovaDelega.page.tsx
```diff
@@ -64,8 +66,10 @@ const NuovaDelega = () => {
   const handleSearchStringChangeInput = useSearchStringChangeInput();
   const [senderInputValue, setSenderInputValue] = useState('');
   const { DELEGATIONS_TO_PG_ENABLED } = getConfiguration();
+  const [delegationCreationSubmitted, setDelegationCreationSubmitted] = useState(false);
 
   const handleSubmit = (values: NewDelegationFormProps) => {
+    setDelegationCreationSubmitted(true);
     void dispatch(createDelegation(values));
```
AndreaCimini90: Instead of using a useState that can cause a useless rerendering, use the then and the catch block of the dispatch method

packages/pn-personagiuridica-webapp/src/pages/Notifiche.page.tsx
```diff
@@ -131,7 +190,10 @@ const Notifiche = ({ isDelegatedPage = false }: Props) => {
     <LoadingPageWrapper isInitialized={pageReady}>
       <Box p={3}>
         {userHasAdminPermissions && !organizationGroup && !isDelegatedPage && (
-          <DomicileBanner source={ContactSource.HOME_NOTIFICHE} />
+          <DomicileBanner
+            source={ContactSource.HOME_NOTIFICHE}
+            onBannerResolved={handleDomicileBannerResolved}
```
AndreaCimini90, line 157: What are we tryind to accomplish here?

AndreaCimini90 (changes requested):

packages/pn-personagiuridica-webapp/src/pages/Notifiche.page.tsx
mflauti: We need this callback to know which domicile banner is actually shown on the page.
The `SEND_PG_YOUR_NOTIFICATION` event requires the `banner` property when a `DomicileBanner` is visible.
Since the banner type is resolved inside the `DomicileBanner` component, the parent page needs a way to receive that value and include it in the Mixpanel payload.

packages/pn-personagiuridica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -58,6 +59,8 @@ import {
 import { resetState } from '../redux/notification/reducers';
 import { RootState } from '../redux/store';
 import { getConfiguration } from '../services/configuration.service';
+import PGEventStrategyFactory from '../utility/MixpanelUtils/PGEventStrategyFactory';
+import { mapNotificationDetailToEventPayload } from '../utility/MixpanelUtils/mappers/notificationPayloadMappers';
```
AndreaCimini90: I think it is better to use the mappers inside the functions that emit events, those defined in MixpanelUtils dir

packages/pn-personagiuridica-webapp/src/components/DomicileBanner/DomicileBanner.tsx
```diff
@@ -98,6 +103,12 @@ const DomicileBanner: React.FC<Props> = ({ source }) => {
     ? null
     : getDomicileData(source, !!defaultSERCQ_SENDAddress, hasCourtesyAddresses, IS_DOD_ENABLED);
 
+  const bannerType = getBannerType(open, domicileBannerData);
```
AndreaCimini90, line 106: Have you also considered using forwardRef and useImperativeHandle instead of using a function and an useEffect?

AndreaCimini90 (changes requested):

packages/pn-personagiuridica-webapp/src/components/DomicileBanner/DomicileBanner.tsx
mflauti: I considered it, but I would avoid `forwardref`/`useImperativeHandle` in this case because the banner type is not really an imperative child API. It is a derived value based on the same inputs used to render the banner.

I think a clearer approach would be to extract the banner type resolution into a pure helper function and let the parent compute the value directly. This way the tracking layer can use the resolved value without relying on child effects or imperative refs.

I can update the impl in this direction if you agree.

packages/pn-personagiuridica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -376,15 +379,33 @@ const NotificationDetail = () => {
...
 
-  /* function which loads relevant information about donwtimes */
+  /* Loads relevant information about downtimes */
   const fetchDowntimeEvents = useCallback((fromDate: string, toDate: string | undefined) => {
     const fetchParams: GetDowntimeHistoryParams = {
       startDate: fromDate,
       endDate: toDate,
     };
-    void dispatch(getDowntimeHistory(fetchParams));
+    void dispatch(getDowntimeHistory(fetchParams))
```
AndreaCimini90: here you can remove void

packages/pn-personagiuridica-webapp/src/pages/Notifiche.page.tsx
```diff
@@ -86,7 +89,20 @@ const Notifiche = ({ isDelegatedPage = false }: Props) => {
       isDelegatedPage,
     };
 
-    void dispatch(getReceivedNotifications(params)).then(() => setPageReady(true));
+    void dispatch(getReceivedNotifications(params))
```
AndreaCimini90: here you can remove void

packages/pn-personagiuridica-webapp/src/pages/NuovaDelega.page.tsx
```diff
@@ -66,7 +68,12 @@ const NuovaDelega = () => {
   const { DELEGATIONS_TO_PG_ENABLED } = getConfiguration();
 
   const handleSubmit = (values: NewDelegationFormProps) => {
-    void dispatch(createDelegation(values));
+    void dispatch(createDelegation(values))
```
AndreaCimini90: here you can remove void

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1899, opened 2026-05-19
feat(PN-19099): use MUI Italia code input and add copy action

gaetano-amato:
## Short description
Update the delegation verification code component according to the MUI Italia 2.0 design system.

- Replaced the custom verification code component with MUI Italia `CodeInput`
- Added copy-to-clipboard action for the verification code
- Applied the change to both Persona Fisica and Persona Giuridica flows
- Added translated tooltip message for successful copy action

## How to test

1. Open the "Nuova Delega" page
2. Create a new delegation request
3. Verify that the verification code is displayed using the MUI Italia `CodeInput` component
4. Click the copy icon next to the verification code
5. Verify that the code is copied to the clipboard
6. Verify that the "Codice copiato" feedback is displayed

**P.S.** Repeat the verification for both Persona Fisica and Persona Giuridica flows

### DIFF
--- a/packages/pn-personagiuridica-webapp/src/components/Deleghe/__test__/VerificationCodeComponent.test.tsx
+++ b/packages/pn-personagiuridica-webapp/src/components/Deleghe/__test__/VerificationCodeComponent.test.tsx
@@ -4,12 +4,12 @@ import VerificationCodeComponent from '../VerificationCodeComponent';
 describe('VerificationCodeComponent', () => {
   it('renders the component and checks the digits', () => {
     const fiveDigits = '12345';
-    const { queryAllByTestId } = render(<VerificationCodeComponent code={fiveDigits} />);
-    const digitsElements = queryAllByTestId('codeDigit');
-    const codes = fiveDigits.split('');
-    expect(digitsElements).toHaveLength(fiveDigits.length);
-    digitsElements.forEach((code, index) => {
-      expect(code).toHaveTextContent(codes[index]);
-    });
+
+    const { container } = render(<VerificationCodeComponent code={fiveDigits} />);
+
+    const input = container.querySelector('input');
+
+    expect(input).toBeInTheDocument();
+    expect(input).toHaveValue(fiveDigits);
   });
 });
--- a/packages/pn-personagiuridica-webapp/src/pages/__test__/NuovaDelega.page.test.tsx
+++ b/packages/pn-personagiuridica-webapp/src/pages/__test__/NuovaDelega.page.test.tsx
@@ -54,7 +54,7 @@ describe('NuovaDelega page', () => {
   });
 
   it('renders the component desktop view', async () => {
-    const { container, getAllByTestId, getByTestId } = render(<NuovaDelega />);
+    const { container, getByTestId } = render(<NuovaDelega />);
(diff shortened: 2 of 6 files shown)


packages/pn-personagiuridica-webapp/src/components/Deleghe/__test__/VerificationCodeComponent.test.tsx
```diff
@@ -2,14 +2,9 @@ import { render } from '../../../__test__/test-utils';
...
-  it('renders the component and checks the digits', () => {
-    const fiveDigits = '12345';
-    const { queryAllByTestId } = render(<VerificationCodeComponent code={fiveDigits} />);
-    const digitsElements = queryAllByTestId('codeDigit');
-    const codes = fiveDigits.split('');
-    expect(digitsElements).toHaveLength(fiveDigits.length);
-    digitsElements.forEach((code, index) => {
-      expect(code).toHaveTextContent(codes[index]);
-    });
+  it('renders the verification code component', () => {
```
AndreaCimini90: Why this change? In the pf webapp we have a similar test and it wasn't change by this pr.
As a general rule avoid to remove checks, but instead try to fix them if they don't work

packages/pn-personagiuridica-webapp/src/pages/__test__/NuovaDelega.page.test.tsx
```diff
@@ -81,11 +81,7 @@ describe('NuovaDelega page', () => {
       'nuovaDelega.form.endDate',
       formatDate(tomorrow.toISOString())
     );
-    const codeDigit = getAllByTestId('codeDigit');
-    const codes = '34153'.split('');
-    codeDigit.forEach((code, index) => {
-      expect(code).toHaveTextContent(codes[index]);
-    });
+    expect(getByTestId('verificationCode')).toBeInTheDocument();
```
AndreaCimini90: Why this change? In the pf webapp we have a similar test and it wasn't change by this pr.
As a general rule avoid to remove checks, but instead try to fix them if they don't work

AndreaCimini90 (changes requested):

packages/pn-personagiuridica-webapp/src/components/Deleghe/__test__/VerificationCodeComponent.test.tsx
gaetano-amato: You're right, thanks. I changed the test because the previous assertions relied on the old codeDigit elements, which are no longer displayed after the switch to MUI Italia CodeInput. I updated both the PF and PG tests to check the actual value of the code displayed by the input, rather than removing the check.

packages/pn-personagiuridica-webapp/src/pages/__test__/NuovaDelega.page.test.tsx
gaetano-amato: You're right, thanks. I changed the test because the previous assertions relied on the old codeDigit elements, which are no longer displayed after the switch to MUI Italia CodeInput. I updated both the PF and PG tests to check the actual value of the code displayed by the input, rather than removing the check.

AndreaCimini90 (approved):

PR #1906, opened 2026-05-28
fix(PN-20069): redirect to login when error is already handled by OneIdentity

alessandrogelmi:
## Short description

Redirect to login when error is already handled by OneIdentity

- Handle OneIdentity errors

## How to test
- Start both PF and PF Login 
- On IDP form click on "Nego il consenso" 
- From the OneIdentity error page click "close" and check that you are redirected directly to Login page (with Mixpanel event)


packages/pn-personafisica-login/src/pages/oneIdentityLoginError/OneIdentityLoginError.tsx
```diff
@@ -22,6 +29,9 @@ const OneIdentityLoginError: React.FC = () => {
     PFLoginEventStrategyFactory.triggerEvent(PFLoginEventsType.SEND_LOGIN_FAILURE, {
```
fbianchicodermine, line 29: if there is no error, do we still send the event?

fbianchicodermine (approved):

PR #1907, opened 2026-05-28
fix(PN-19455): show error on payment if retrieval ID is expired

alessandrogelmi:
## Short description
Show error on payment if retrieval ID is expired

- TPP error handling

## How to test
https://pagopa.atlassian.net/wiki/spaces/PN/pages/2949382203/SRS+Permanenza+accesso+da+terze+parti#In-test


packages/pn-commons/src/components/NotificationDetail/NotificationPaymentRecipient.tsx
```diff
@@ -263,9 +269,26 @@ const NotificationPaymentRecipient: React.FC<Props> = ({
             </Box>
           )}
           {errorOnPayment && (
-            <Alert severity="error" variant="outlined" data-testid="payment-error">
-              {getErrorMessage()}
-            </Alert>
+            <MIAlert
```
AndreaCimini90, line 301: I'll make you a suggestion. Instead of using another state, we could modify errorOnPayment to be an object with a title and description. If it's null, it means there are no errors. If it's defined, it can have a title and description. The title will be optional, while the description will be mandatory. In the case of a standard error, the description will be getErrorMessage.

packages/pn-personafisica-webapp/src/hooks/useDismissToastOnError.ts
```diff
@@ -0,0 +1,14 @@
+import { useCallback, useEffect } from 'react';
+
+import { AppResponse, AppResponsePublisher } from '@pagopa-pn/pn-commons';
+
+import { ServerResponseErrorCode } from '../utility/AppError/types';
+
+export const useDismissToastOnError = (actionId: string, errorCode: ServerResponseErrorCode) => {
+  const handler = useCallback((e: AppResponse) => e.errors?.[0]?.code !== errorCode, [errorCode]);
```
AndreaCimini90: Pass handler as parameter to made this use more flexible

packages/pn-commons/src/components/NotificationDetail/NotificationPaymentRecipient.tsx
```diff
@@ -151,14 +152,19 @@ const NotificationPaymentRecipient: React.FC<Props> = ({
   const handleCheckPaymentSelected = (paymentType: 'default' | 'tpp') => {
     if (selectedPayment.pagoPa) {
       setErrorOnPayment(false);
+      setTppPaymentError(false);
       if (paymentType === 'tpp') {
-        onPayTppClick?.(
+        const result = onPayTppClick?.(
```
AndreaCimini90: To manage the error we can use the hook just created, but the component must accept the action as property

AndreaCimini90 (changes requested):

packages/pn-personafisica-webapp/tsconfig.json
```diff
@@ -23,6 +23,10 @@
     "noFallthroughCasesInSwitch": true,
     "resolveJsonModule": true
   },
-  "include": ["src/**/*"],
+  "include": [
+    "src/**/*",
+    "../pn-commons/src/hooks/useDismissToastOnError.ts",
```
AndreaCimini90: why do we need this?

AndreaCimini90 (changes requested):

packages/pn-personafisica-webapp/tsconfig.json
alessandrogelmi: removed

AndreaCimini90 (dismissed):

AndreaCimini90 (approved):

PR #1910, opened 2026-06-02
Feat(PN-19761): mixpanel event tracking (PG)

mflauti:
## Short description
Complete PG mixpanel tracking implementation according to CX event specifications.

- implement PG Mixpanel events and super properties
- add dedicated payload mappers where payload transformation is required
- move payload mapping from UI components/pages to mixpanel tracking layer
- add/update tests for PG tracking configs and mappers.

## How to test
Run PG web app and verify the implemented events are triggered properly


packages/pn-personagiuridica-webapp/src/models/PGEventsType.ts
```diff
@@ -1,9 +1,63 @@
...
+  SEND_PG_TIMELINE_DOWNLOAD = 'SEND_PG_TIMELINE_DOWNLOAD',
+
+  /* SERVICE STATUS */
+  SEND_PG_SERVICE_STATUS = 'SEND_PG_SERVICE_STATUS',
+
+  /* SUPER PROPERTIES */
+  USER_ROLE = 'USER_ROLE',
+  SEND_PG_HAS_EMAIL = 'SEND_PG_HAS_EMAIL',
+  SEND_PG_HAS_SMS = 'SEND_PG_HAS_SMS',
+  SEND_PG_HAS_DIGITAL_DOMICILE = 'SEND_PG_HAS_DIGITAL_DOMICILE',
```
fbianchicodermine, line 59: I can't find event `SEND_PG_HAS_DIGITAL_DOMICILE` in the documentation.

Maybe it's `SEND_PG_HAS_SERCQ`? (super property with a yes/no value)

fbianchicodermine (changes requested):

packages/pn-personagiuridica-webapp/src/models/PGEventsType.ts
mflauti: Yes, `SEND_PG_HAS_DIGITAL_DOMICILE` is intentional.
The event was initially defined as `SEND_PG_HAS_SERCQ`, with a `yes/no` value. After some follow-up clarifications with CX, the requirement was changed to a more general super property representing the configured digital domicile type rather than only indicating whether SERCQ is set.

For this reason the final event is `SEND_PG_HAS_DIGITAL_DOMICILE`, with values `PEC`, `SERCQ_SEND` or `not_available`.

It looks like the corresponding Airtable documentation was not updated after this change, I'll ask CX to align it with the final requirement.

fbianchicodermine (approved):

PR #1911, opened 2026-06-03
feat(PN-19760): Add PA mixpanel tracking foundation and events

mflauti:
## Short description
Complete PA mixpanel tracking implementation according to CX event specifications.

- Add PA mixpanel tracking foundation based on PG architecture
- Implement PA mixpanel events
- Add dedicated payload mappers where payload transformation is required
- Add tests

## How to test
Run PA web app and verify the implemented events are triggered with the expected payloads.


packages/pn-pa-webapp/src/pages/NewNotification.page.tsx
```diff
@@ -72,6 +74,28 @@ const NewNotification = () => {
...
+      return;
+    }
+    if (IS_PAYMENT_ENABLED && step === 3) {
+      PAEventStrategyFactory.triggerEvent(PAEventsType.SEND_PA_DEBT_POSITION_DETAIL);
+      return;
+    }
+    if ((IS_PAYMENT_ENABLED && step === 4) || (!IS_PAYMENT_ENABLED && step === 2)) {
+      PAEventStrategyFactory.triggerEvent(PAEventsType.SEND_PA_DOCUMENTATION);
+    }
+  };
```
alessandrogelmi: You can move this function outside the component and perhaps use a map with the step as the key and the event as the value to avoid all those if statements

packages/pn-pa-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -140,6 +146,7 @@ const NotificationDetail: React.FC = () => {
   };
 
   const legalFactDownloadHandler = (legalFact: LegalFactId) => {
+    PAEventStrategyFactory.triggerEvent(PAEventsType.SEND_PA_TIMELINE_DOWNLOAD, { legalFact });
```
alessandrogelmi, line 149: Why are you passing the legalFact? On Airtable the `SEND_PA_TIMELINE_DOWNLOAD` event has no properties

alessandrogelmi (changes requested):

packages/pn-pa-webapp/src/pages/NotificationDetail.page.tsx
mflauti: I added `document_type` because I mirrored the equivalent PG event, `SEND_PG_TIMELINE_DOWNLOAD`, which sends the legal fact category through this property.

However, I notied that the PA specification does not currently include it, so I'm checking with CX whether it should be kept or removed.

mflauti: CX just confirmed that the missing document_type property in Airtable was an oversight and that it is expected for SEND_PA_TIMELINE_DOWNLOAD, as already defined in Figma.

I'll therefore keep the current implementation.

alessandrogelmi (approved):

PR #1930, opened 2026-06-18
fix(A2-2339): improve mobile contact accordion a11y

mflauti:
## Short description
This PR improves the mobile a11y of the contact cards accordion used in PF and PG contact pages.

- replace the mobile expand/collapse icon-only behavior with an accessible accordion header
- add proper accordion semantics and keyboard support according to the WAI-ARIA accordion pattern
- keep contact cards static on desktop and when the contact is already active
- refactor contact card actions so mobile menu items can be triggered using keyboard
- align PF and PG contact cards with the updated `PnInfoCard` API
- update tests

## How to test
1. open the PF and PG contact pages on a mobile viewport
2. using keyboard navigation and a screen reader, verify that inactive contact cards are announced as expandable/collapsible controls and that the state is read correctly
3. verify that the accordion headers can be focused and toggled with both `Enter` and `Space`
4. verify that active contact cards are not announced as accordions and remain always expanded
5. open the three-dots menu on active contacts and verify, using keyboard navigation and a screen reader, that the available actions are correctly announced and can be triggered using both `Enter` and `Space`
6. verify that the desktop behavior is unchanged: contact cards are always expanded and actions are shown as regular buttons.


packages/pn-commons/src/components/PnInfoCard.tsx
```diff
@@ -79,91 +107,202 @@ const PnInfoCardActions: React.FC<Pick<Props, 'actions'>> = ({ actions }) => {
...
+              transform: expanded ? 'rotate(180deg)' : 'rotate(0deg)',
+              transition: theme.transitions.create('transform', {
+                duration: theme.transitions.duration.shortest,
+              }),
+            })}
+          />
+        </ButtonBase>
+      </Box>
+
+      <Collapse in={expanded} timeout="auto">
```
AndreaCimini90: Before the content of a card was removed from DOM if collapsed. Now is always mounted. This can leads to unexpected side effects. Try to use unmountOnExit

AndreaCimini90 (changes requested):

packages/pn-commons/src/components/PnInfoCard.tsx
mflauti: You're right! I kept the panel mounted to keep the aria-controls target available, but this changes the content lifecycle.. I'll use mountOnEnter and unmountOnExit so the content is mounted only when expanded and removed again after collapse.

AndreaCimini90 (approved):

AndreaCimini90 (changes requested):
The desktop layout is changed:

Now

Image

Before

Image

AndreaCimini90 (approved):

PR #1942, opened 2026-07-06
fix(PN-20241): update empty states UI for recipient users

alessandrogelmi:
## Short description
Update empty states UI for recipient users

- Update UI
- Update contents

## How to test
On PF and PG check the UI of empty states with: filters applied, delegated user and default state


packages/pn-personagiuridica-webapp/src/components/Notifications/DesktopNotifications.tsx
```diff
@@ -1,12 +1,9 @@
 import { useRef } from 'react';
```
AndreaCimini90, line 1: It's missed the updates on MobileNotifications.tsx

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #1945, opened 2026-07-08
feat(PN-20560): Add FIMS token exchange flow to PF

mflauti:
## Short description
Adds the frontend FIMS token exchange flow for PF.
When the user lands on PF with a `fimsToken` in the URL hash, the `SesionGuard` exchanges it through `/fims/exchange` and creates the SEND session.

### DIFF
--- a/packages/pn-personafisica-webapp/src/navigation/SessionGuard.tsx
+++ b/packages/pn-personafisica-webapp/src/navigation/SessionGuard.tsx
@@ -108,22 +108,6 @@ const SessionGuard = () => {
     }
   };
 
-  const removeFimsTokenFromHash = () => {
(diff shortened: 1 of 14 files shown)


packages/pn-personafisica-webapp/src/navigation/SessionGuard.tsx
```diff
@@ -146,7 +175,10 @@ const SessionGuard = () => {
   };
 
   useEffect(() => {
-    if (spidToken) {
+    if (fimsToken) {
+      removeFimsTokenFromHash();
```
fbianchicodermine: it is not needed

fbianchicodermine (changes requested):

fbianchicodermine (approved):

PR #1949, opened 2026-07-13
fix(PN-20645): fix link in footer

fbianchicodermine:
## Short description
Fix broken link "Sicurezza delle informazioni" in footer

## How to test
Start an app (only PFL is up to SEND) and check that link "Sicurezza delle informazioni" exists

### Note:
[X] - PF -> ok è in carico a SEND
[ ] - PG -> è in carico a selfcare
[ ] - PA -> è in carico a OneIdentity


SarahDonvito (commented):
Warning: this fix only concerns PF login because the label in footer is present only for before login and SEND manage only PF login. PG and PA login are managed by other team.

SarahDonvito (approved):

PR #1958, opened 2026-07-23
fix(PN-20781): hide "new attachment" timeline message when the analog event has no attachments

fbianchicodermine:
## Short description
Fix the notification timeline showing `"C'è un nuovo documento allegato."` on `SEND_ANALOG_PROGRESS` events that have no attachments.
With attachments → still shows "C'è un nuovo documento allegato." (unchanged).
Without attachments → falls back to the existing generic message "C'è un aggiornamento sull'invio cartaceo.".
All other delivery codes are unaffected.

- `packages/pn-commons/src/utility/TimelineUtils/SendAnalogFlowStep.ts`: add the attachment-code list and the attachments guard.
- `packages/pn-commons/src/utility/TimelineUtils/__test__/SendAnalogFlowStep.test.ts`: 2 new tests (code with/without attachments).

## How to test
Log in PF dev with "fieramosca" and check `GZLZ-MQJM-EKLH-202607-M-1` invalid event in delivering status

### DIFF
--- a/packages/pn-commons/src/utility/TimelineUtils/SendAnalogFlowStep.ts
+++ b/packages/pn-commons/src/utility/TimelineUtils/SendAnalogFlowStep.ts
@@ -25,6 +25,27 @@ const SEND_ANALOG_FEEDBACK_KO_DETAIL_CODES = [
   'RECRI004C',
 ];
 
+// CODES with "C'è un nuovo documento allegato." description
+const SEND_ANALOG_FLOW_NEW_ATTACHMENT_DETAIL_CODES = [
+  'RECRN001B',
+  'RECRN002B',
+  'RECRN002E',
+  'RECRN003B',
+  'RECRN004B',
+  'RECRN005B',
+  'RECAG001B',
+  'RECAG002B',
+  'RECAG003B',
+  'RECAG003E',
+  'RECAG005B',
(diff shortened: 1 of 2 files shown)


packages/pn-commons/src/utility/TimelineUtils/SendAnalogFlowStep.ts
```diff
@@ -25,6 +25,27 @@ const SEND_ANALOG_FEEDBACK_KO_DETAIL_CODES = [
   'RECRI004C',
 ];
 
+// CODES (PF,PG,PA) with description: "C'è un nuovo documento allegato." 
+const SEND_ANALOG_FLOW_NEW_ATTACHMENT_DETAIL_CODES = [
```
AndreaCimini90, line 29: Do we really need this list? Wouldn't the condition of whether or not there are attachments be enough?

fbianchicodermine: The wrong text doesn't come from the attachments but from i18n key mapping.
`getTimelineStepInfo` runs for every event, so suppressing the description whenever attachments are empty would also hide those legitimate descriptions

AndreaCimini90 (approved):

PR #1962, opened 2026-07-23
feat(PN-20240): Update PA notification detail layout for legal communications

mflauti:


packages/pn-commons/src/components/NotificationDetail/NotificationDetailSection.tsx
```diff
@@ -26,9 +32,25 @@ const NotificationDetailSection = ({
   isDelegate,
   isLessThan10Years,
   downloadFilesMessage,
+  showAllAarDocuments = false,
```
AndreaCimini90: I don't understand why we need this. In case of recipient we have only one AAR and in case of PA depends on the number of the recipient. So showAllAarDocuments should be always true

packages/pn-commons/src/components/NotificationDetail/NotificationDetailSection.tsx
```diff
@@ -26,9 +32,25 @@ const NotificationDetailSection = ({
...
+  useDocumentTitle = false,
```
AndreaCimini90: I don't understand why we need this.

packages/pn-commons/src/components/NotificationDetail/NotificationDetailSection.tsx
```diff
@@ -81,54 +103,63 @@ const NotificationDetailSection = ({
...
-              </Typography>
-            </Stack>
-            {!isCancelled && (
-              <MIIconButton
-                data-testid="documentButton"
-                edge="end"
-                aria-label={getLocalizedOrDefaultLabel(
-                  'notifications',
-                  'detail.notification-detail-section.aria-label'
+                <Stack>
```
AndreaCimini90, line 121: If the notification is multirecipient we should show the denomination and the fiscal code near the AAR label, as it was before the rework.

Image

packages/pn-pa-webapp/public/locales/en/notifiche.json
```diff
@@ -25,12 +26,11 @@
   },
   "payment": {
     "title": "Payments",
-    "subtitle-single": "Here, you can track the payment status of the pagoPA alerts attached to the notification.",
-    "subtitle-single-f24": "Here, you can see the F24 forms attached to the notification.",
-    "subtitle-multiple": "To see the payment status of the pagoPA alerts attached to the notification, select a recipient.",
```
AndreaCimini90, line 30: Sorry my fault! I should leave the content subtitle-multiple and subtitle-multiple-f24

packages/pn-pa-webapp/src/components/Notifications/NotificationRecipientsDetail.tsx
```diff
@@ -1,97 +0,0 @@
-import { useState } from 'react';
```
AndreaCimini90, line 1: What this component did?

AndreaCimini90 (changes requested):

packages/pn-commons/src/components/NotificationDetail/NotificationDetailSection.tsx
mflauti: You're right. I introduced this prop while refactoring the original single-AAR implementation done for PF/PG to adapt it to PA requirements, where the component has to iterate over the AAR documents in case of multi-recipient notifications. I used it to preserve the previous PF/PG behavior, where the component rendered only the first matching AAR. However, PF/PG already restrict the data to the current recipinet, so this additional restriction is not needed.

packages/pn-commons/src/components/NotificationDetail/NotificationDetailSection.tsx
mflauti: You're right, `useDocumentTitle` was not needed. I introduced it while adapting the component to PA. I'll remove it!

packages/pn-pa-webapp/src/components/Notifications/NotificationRecipientsDetail.tsx
mflauti: This was the old PA detail table content used to show notification details in the previous layout. Its responsibility has been replaced by the new summary card (based on AbstractPaper) plus NotificationDetailDrawer, so it is no more needed.

AndreaCimini90 (approved):

PR #1969, opened 2026-07-28
fix(PN-20848): Add missing static sections to PF informal notification details

gaetano-amato:
## Summary

This PR fixes the rendering of the static content in the **Persona Fisica (PF)** and **Persona Giuridica (PG)**  informal notification detail page

### Changes
- Added greeting
- Added attachments informational section
- Added payment informational section
- Added assistance informational section

### Tested
- Informal notification with attachments
- Informal notification with payment
- Informal notification without attachments/payment
- languages localization


packages/pn-commons/src/components/NotificationDetail/AbstractPaper.tsx
```diff
@@ -190,9 +212,67 @@ const AbstractPaper = ({
                   </Typography>
                 </>
               ) : (
-                <Box sx={{ overflowWrap: 'anywhere' }}>
-                  <PNMarkdown content={abstract ?? ''} />
-                </Box>
+                <Stack>
+                  {recipientDenomination && (
+                    <Typography variant="body1" color="text.primary">
+                      <Trans
```
AndreaCimini90: Trans is used when we have to render components inside the string, as a strong tag or a more complex component. If the string have only text and some dynamic part, we can use the getLocalizedOrDefaultLabel passing as forth parameter the dynamic data. For example the string "Hi, my name is {{name}}" can be translated with "getLocalizedOrDefaultLabel(namespace, path, undefined, {name: "Andrea"})"

packages/pn-commons/src/utility/localization.utility.ts
```diff
@@ -57,3 +57,10 @@ export function getLocalizedOrDefaultLabel(
   }
   return defaultLabel ?? path;
 }
+
+export function getTranslationMessage(key: string): { key: string; ns: string } {
```
AndreaCimini90: If it is ment to bee a common function, maybe the ns should be passed as parameter

AndreaCimini90 (changes requested):

AndreaCimini90 (dismissed):

AndreaCimini90 (approved):

PR #1975, opened 2026-07-31
fix(PN-20854): resolve PA notification detail issues

mflauti:
## Short description
Fixes several PA notification detail issues

Refs: PN-20854, PN-20857, PN-20904, PN-20925

- fix notification status content in the detail aside
- fix unavailable documents and AAR messages
- fix recipients rendering in notification summary and drawer
- fix mobile order of notification detail sections
- fix payment notice code and status chip spacing
- fix legal disclaimer visibility when abstract is missing
- remove duplicated Vite config type import


packages/pn-commons/src/components/NotificationDetail/AbstractPaper.tsx
```diff
@@ -200,81 +200,78 @@ const AbstractPaper = ({
               </Box>
             </Grid>
           </Grid>
-          {abstract && (
```
AndreaCimini90, line 203: This should be restored

packages/pn-commons/src/components/NotificationDetail/AbstractPaper.tsx
```diff
@@ -200,81 +200,78 @@ const AbstractPaper = ({
...
-                      overflowWrap: 'anywhere',
-                      '& p': {
-                        m: 0,
-                        typography: 'body1',
-                        color: 'text.primary',
-                        mt: 4,
-                      },
-                    }}
-                  >
-                    <PNMarkdown content={abstract} />
```
AndreaCimini90, line 239: if we restored the  `{abstract && ( ...` here we can avoid the `?? ''`

packages/pn-pa-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -365,6 +346,43 @@ const NotificationDetail: React.FC = () => {
     },
   ].filter((detail) => detail.value);
 
+  const notificationAside = (
```
AndreaCimini90: Are you sure we need this? We can solve with the order property without using `{!isMobile && notificationAside}` && `{isMobile && notificationAside}`?

AndreaCimini90 (changes requested):

packages/pn-commons/src/components/NotificationDetail/AbstractPaper.tsx
mflauti: Do you mean restoring that condition only around the element that renders the abstract?
Previously it also wrapped the legal disclaimer, but the disclaimer must always be shown.
If so, I agree.

packages/pn-commons/src/components/NotificationDetail/AbstractPaper.tsx
mflauti: Yes, if we change the code so that the `PNMarkdown` is rendered only inside the abstract guard we can remove the `?? ''`

packages/pn-pa-webapp/src/pages/NotificationDetail.page.tsx
mflauti: I already tried a similar approach working on this fix, but it can detach the cancel buttom from the main content when the vertical space occupied by the aside is greater than the main content area. In that case the button is rendered below the aside content instead of staying visually connected to the main notification content.

AndreaCimini90 (approved):

PR #1976, opened 2026-07-31
fix(PN-20950): show F24 section when there are only payments of this type

AndreaCimini90:
## Short description
When a notification has only f24 payments, the section is not shown. This pr fixes this bug.

- Changed the condition in NotificationPaymentRecipient

## How to test
Login with PF, user franco -> check that for notification with pagoPA payments, notification with pagoPA + F24 payments and notification with F24 payments, the section is correctly shown

### DIFF
--- a/packages/pn-commons/src/components/NotificationDetail/NotificationPaymentRecipient.tsx
+++ b/packages/pn-commons/src/components/NotificationDetail/NotificationPaymentRecipient.tsx
@@ -311,7 +311,7 @@ const NotificationPaymentRecipient: React.FC<Props> = ({
         </>
       )}
 
-      {!isCancelled && f24Only.length > 0 && (
+      {!isCancelled && f24Only.length > 0 && pagoPaF24.length > 0 && (
         <Divider
           component="div"
           role="presentation"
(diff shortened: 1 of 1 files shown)


packages/pn-commons/src/components/NotificationDetail/NotificationPaymentRecipient.tsx
```diff
@@ -321,7 +321,7 @@ const NotificationPaymentRecipient: React.FC<Props> = ({
         </Divider>
       )}
 
-      {!isCancelled && f24Only.length > 0 && pagoPaF24.length > 0 && (
+      {!isCancelled && f24Only.length > 0 && (
         <>
           {f24Only.length > 0 && pagoPaF24.length > 0 && (
```
fbianchicodermine: Leftover redundant check: now that the outer block is guarded by `f24Only.length > 0`, testing it again here is dead code. This can be reduced to:

```suggestion
          {pagoPaF24.length > 0 && (
```

fbianchicodermine (changes requested):

fbianchicodermine (approved):

PR #1977, opened 2026-07-31
fix(PN-20939): show onboarding when the unread notifications are only those in EFFECTIVE_DATE

AndreaCimini90:
## Short description
If we have an user that have notifications to read, but those notifications are in status EFFECTIVE_DATE, the Onboarding must be shown.

- Changed the condition in OnboardingGuard, so notifications in status EFFECTIVE_DATE doesn't affect the Onboarding
- Added isFreshLogin check when login with OI

## How to test
Login with PA -> simulate with netify a call to the notification list where the notifications to read (isNewNotification = true) are only those in status EFFECTIVE_DATE

### DIFF
--- a/packages/pn-personafisica-webapp/src/navigation/OnboardingGuard.tsx
+++ b/packages/pn-personafisica-webapp/src/navigation/OnboardingGuard.tsx
@@ -17,7 +17,7 @@ import * as routes from './routes.const';
 
 const hasNotificationsToRead = (notifications: Array<RecipientNotification>): boolean =>
   notifications.some(
-    (n) => n.isNewNotification || n.notificationStatus === NotificationStatus.EFFECTIVE_DATE
+    (n) => n.isNewNotification || n.notificationStatus !== NotificationStatus.EFFECTIVE_DATE
   );
 
 const OnboardingGuard: React.FC = () => {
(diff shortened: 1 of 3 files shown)


fbianchicodermine (changes requested):
The fix itself looks right, but the test suite is red on this branch.

`'does not redirect when user has EFFECTIVE_DATE notifications'` (`OnboardingGuard.test.tsx:159`) encodes exactly the previous behaviour: a notification with `isNewNotification: false` in `EFFECTIVE_DATE` was expected to *suppress* the onboarding. With the new condition that notification no longer counts as "to read", so the guard redirects and the assertion fails:

```
Test Files  1 failed (1)
     Tests  1 failed | 7 passed (8)
 ❯ src/navigation/__test__/OnboardingGuard.test.tsx:180
   expect(screen.queryByText('Onboarding Page')).toBeInTheDocument()
```

The test needs to be inverted and renamed (e.g. `redirects to onboarding when the only unread notifications are in EFFECTIVE_DATE`), expecting `routes.ONBOARDING`.

While you're there, two cases worth adding, since today neither the ticket scenario nor its inverse is covered:
- `isNewNotification: true` + `EFFECTIVE_DATE` → onboarding shown. This is the actual PN-20939 scenario; the existing test uses `isNewNotification: false`, which is a different case.
- mixed list: one unread in `EFFECTIVE_DATE` **and** one unread in another status → no redirect. This is the one that guards against the condition being flipped back.

`SessionGuard.test.tsx` is green (19/19).

fbianchicodermine (approved):

PR #1979, opened 2026-07-31
fix(PN-20849): resolve PA notification list issues

mflauti:
## Short description
Fixes the PA sent notifications list layout and labels on desktop/mobile, aligning the mobile card content, header actions, spacing and recipient column copy with the expected design.

Refs: PN-20839, PN-20849

- fix recipient column label in sent notifications list
- align mobile notification card fields with expected labels and status placement
- align mobile header actions, subtitle and filter entry point
- add filter icon support to `CustomMobileDialogToggle`


packages/pn-pa-webapp/src/pages/Dashboard.page.tsx
```diff
@@ -43,6 +43,7 @@ const Dashboard = () => {
   const isSupportUser = useAppSelector(authSelectors.selectIsSupportUser);
   const navigate = useNavigate();
   const isMobile = useIsMobile();
+  const isSmallScreen = useIsMobile('sm');
```
AndreaCimini90, line 46: We can use isMobile. Because all the mobile view is based on isMobile flag

AndreaCimini90 (changes requested):

packages/pn-pa-webapp/src/pages/Dashboard.page.tsx
mflauti: `isMobile` uses the default `lg` breakpoint, so it becomes true below 1200px.
For the notifications list this is appropriate, because the table is replaced by the mobile cards at that breakpoint.
For the page actions, however, using the same flag would stack the two buttons and make them fullwidth already on teblet/intermediate viewport sizes, where there's still enough space to keep the desktop layout.
As shown in the attached screenshot, this results in unnecessarly large buttons.
For this reason I used `useIsMobile('sm')` for the header actions, while `isMobile` remains responsible for switching the notifications list.
image

AndreaCimini90: ok you're right

AndreaCimini90 (approved):

PR #1990, opened 2026-08-05
feat(PN-20886): implemented data lake maintenance alert

AndreaCimini90:
## Short description
When data lake is under maintenance an alert should be shown in the statistics page.

- Added new function in date.utility that check if a date is in a range
- Added new configuration to set the date range when the alert must be shown
- Added new localizations for the alert title and content -> the phrase "e open data" must be removed from the description
- Added the alert to the Statistics.page

## How to test
- Add new configuration DATA_LAKE_MAINTENANCE_DATES in your config.json
- Set the value of DATA_LAKE_MAINTENANCE_DATES as `[startDate]_[endDate]` where startDate and endDate are in the format YYYY-MM-DD or YYYY-MM-DDTHH:mm:ssZ or YYYY-MM-DDTHH:mm:ss:ssss
- Login with PA -> go to statistics -> check that the alert is shown.

N.B. to simulate the date you can set dates near the moment you are testing or use a chrome extension that simulate time. In this second case you have to modify the expiration of the token, otherwise you will be logged out

### DIFF
--- a/packages/pn-commons/src/utility/__test__/date.utility.test.ts
+++ b/packages/pn-commons/src/utility/__test__/date.utility.test.ts
@@ -290,9 +290,26 @@ describe('Date utility', () => {
     futureDateDays.setDate(now.getDate() + 7);
 
     const futureDateMonth = new Date(now);
-    futureDateMonth.setDate(now.getMonth() + 1);
+    futureDateMonth.setMonth(now.getMonth() + 1);
 
     const result = isDateInRange(now, futureDateDays.toISOString(), futureDateMonth.toISOString());
+    expect(result).toBeFalsy();
+  });
+
+  it('isDateInRange - returns true when date is not in range (no time specified)', () => {
+    const now = new Date();
+
+    const futureDateDays = new Date(now);
+    futureDateDays.setDate(now.getDate() + 7);
+
+    const futureDateMonth = new Date(now);
+    futureDateMonth.setMonth(now.getMonth() + 1);
+
+    const result = isDateInRange(
+      now,
+      formatToSlicedISOString(futureDateDays),
+      formatToSlicedISOString(futureDateMonth)
+    );
     expect(result).toBeFalsy();
   });
 });
--- a/packages/pn-commons/src/utility/__test__/date.utility.test.ts
+++ b/packages/pn-commons/src/utility/__test__/date.utility.test.ts
@@ -18,6 +18,7 @@ import {
   getEndOfDay,
   getStartOfDay,
   getWeeksFromDateRange,
+  isDateInRange,
   isToday,
   minutesBeforeNow,
   subtractMonthsFromDate,
@@ -267,5 +268,31 @@ describe('Date utility', () => {
 
     const result = clampMax(dateBefore, max);
     expect(result).toBe(dateBefore);
+  });
+
+  it('isDateInRange - returns true when date is in range', () => {
+    const now = new Date();
+
+    const futureDateDays = new Date(now);
+    futureDateDays.setDate(now.getDate() + 7);
+
+    const pastDateDays = new Date(now);
+    pastDateDays.setDate(now.getDate() - 5);
+
+    const result = isDateInRange(now, pastDateDays.toISOString(), futureDateDays.toISOString());
+    expect(result).toBeTruthy();
+  });
+
+  it('isDateInRange - returns false when date is not in range', () => {
+    const now = new Date();
+
+    const futureDateDays = new Date(now);
+    futureDateDays.setDate(now.getDate() + 7);
+
+    const futureDateMonth = new Date(now);
+    futureDateMonth.setDate(now.getMonth() + 1);
+
+    const result = isDateInRange(now, futureDateDays.toISOString(), futureDateMonth.toISOString());
+    expect(result).toBeFalsy();
   });
 });
--- a/packages/pn-commons/src/utility/date.utility.ts
+++ b/packages/pn-commons/src/utility/date.utility.ts
@@ -244,3 +244,33 @@ export function subtractMonthsFromDate(
 export function clampMax(date: Date, max: Date): Date {
   return date.getTime() > max.getTime() ? max : date;
 }
+
+/**
+ * Checks if the date is within a date range
+ *
+ * @param {Date} date current date
+ * @param {string} startDate start date in format YYYY-MM-DD
+ * @param {string} endDate end date in format YYYY-MM-DD
+ * @returns {boolean} True if the date is in the date range, false otherwise
+ */
+export const isDateInRange = (date: Date, startDate: string, endDate: string): boolean => {
+  const current = new Date(date);
+
+  // Note: Month is 0-based in JavaScript Date constructor (August = 7)
+  // const startDate = new Date(2026, 7, 14, 0, 0, 0);
+  // const endDate = new Date(2026, 7, 25, 23, 59, 59);
+  const start = new Date(startDate);
+  // If startDate does NOT contain 'T', set it to the beginning of the day (00:00:00.000)
+  if (!startDate.includes('T')) {
+    start.setHours(0, 0, 0, 0);
+  }
+
+  // at the end of the day
+  const end = new Date(endDate);
+  // If endDate does NOT contain 'T', set it to the end of the day (23:59:59.999)
+  if (!endDate.includes('T')) {
+    end.setHours(23, 59, 59, 999);
+  }
+
+  return current >= start && current <= end;
+};
--- a/packages/pn-commons/src/utility/date.utility.ts
+++ b/packages/pn-commons/src/utility/date.utility.ts
@@ -1,4 +1,4 @@
-import { add, addDays, compareAsc } from 'date-fns';
+import { add, addDays, compareAsc, parseISO } from 'date-fns';
 
 import DateFnsAdapter from '@date-io/date-fns';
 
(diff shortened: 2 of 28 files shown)


packages/pn-commons/src/utility/__test__/date.utility.test.ts
```diff
@@ -268,4 +269,30 @@ describe('Date utility', () => {
...
+  });
+
+  it('isDateInRange - returns false when date is not in range', () => {
+    const now = new Date();
+
+    const futureDateDays = new Date(now);
+    futureDateDays.setDate(now.getDate() + 7);
+
+    const futureDateMonth = new Date(now);
+    futureDateMonth.setDate(now.getMonth() + 1);
```
fbianchicodermine: `setDate` sets the day of the month, but the argument here is a month index: `getMonth() + 1` is `8` in August, so `futureDateMonth` becomes **8 Aug**, not 5 Sep.

The range passed to `isDateInRange` is therefore inverted (`start = now + 7d` > `end`), and the assertion passes because `start > end` — not because the current date precedes a future window, which is the case this test is named after and which nothing else in the suite covers. It also can never fail: `getMonth() + 1` is always 1–12, so `futureDateMonth` always lands in the first 12 days of the current month, always before `now + 7d`.

```suggestion
    futureDateMonth.setMonth(now.getMonth() + 1);
```

packages/pn-commons/src/utility/__test__/date.utility.test.ts
```diff
@@ -268,4 +269,30 @@ describe('Date utility', () => {
...
+  it('isDateInRange - returns true when date is in range', () => {
+    const now = new Date();
+
+    const futureDateDays = new Date(now);
+    futureDateDays.setDate(now.getDate() + 7);
+
+    const pastDateDays = new Date(now);
+    pastDateDays.setDate(now.getDate() - 5);
+
+    const result = isDateInRange(now, pastDateDays.toISOString(), futureDateDays.toISOString());
```
fbianchicodermine, line 282: Both tests pass `.toISOString()`, which always contains `T`, so the `!includes('T')` branch of `isDateInRange` is never executed. That branch is exactly the one triggered by the real config value (`"2026-08-14_2026-08-25"` in `config-dev.json`): the only non-trivial logic in the function is the only part not covered.

Can we add a date-only case, e.g. `isDateInRange(new Date('2026-08-20T12:00:00'), '2026-08-14', '2026-08-25')` → `true`, plus the two boundary days (14th at 00:00 and 25th at 23:59 should both be inside)?

packages/pn-commons/src/utility/date.utility.ts
```diff
@@ -244,3 +244,33 @@ export function subtractMonthsFromDate(
...
+  if (!startDate.includes('T')) {
+    start.setHours(0, 0, 0, 0);
+  }
+
+  // at the end of the day
+  const end = new Date(endDate);
+  // If endDate does NOT contain 'T', set it to the end of the day (23:59:59.999)
+  if (!endDate.includes('T')) {
+    end.setHours(23, 59, 59, 999);
+  }
```
fbianchicodermine: `new Date('2026-08-14')` is parsed as **UTC** midnight (per the ES spec, date-only forms are UTC), but `setHours` writes to the **local** components. At UTC+2 the result is correct by coincidence; at a negative offset it is off by one day — at UTC-5 `new Date('2026-08-14')` is already 13 Aug locally, so `setHours(0, 0, 0, 0)` pins it to 13 Aug 00:00 and the whole window shifts a day earlier. Same for `end`.

Practical impact is low for an Italian PA, but the `includes('T')` heuristic can go away entirely by using `parseISO` (which treats date-only forms as local) together with the helpers already in this file at lines 63-69:

```suggestion
  const start = startDate.includes('T') ? parseISO(startDate) : getStartOfDay(parseISO(startDate));
  const end = endDate.includes('T') ? parseISO(endDate) : getEndOfDay(parseISO(endDate));
```

This needs `parseISO` added to the `date-fns` import. With this, `const current = new Date(date)` at line 257 also becomes redundant — `date` is already a `Date` and can be compared directly.

packages/pn-pa-webapp/src/pages/__test__/Statistics.page.test.tsx
```diff
@@ -312,6 +330,22 @@ describe('Statistics Page tests', () => {
     });
   });
 
+  it('show maintenance alert', async () => {
+    statisticsUnderMaintenance = true;
```
fbianchicodermine: `statisticsUnderMaintenance` is never reset, so every test declared after this one runs with the maintenance window active. The suite stays green only because nothing asserts the alert is absent — it's a landmine for the next test added to this file. Can we add `statisticsUnderMaintenance = false;` to the existing `afterEach` at line 70?

Related: the mock at line 39 replaces `getConfiguration` with an object holding only `STATISTICS_MAINTENANCE_DATES`, for the whole file, dropping every other config key. Nothing under `Statistics.page` reads config today, so it works, but spreading the real one would be sturdier:

```ts
getConfiguration: () => ({ ...actual.getConfiguration(), STATISTICS_MAINTENANCE_DATES: ... })
```

Alternatively, drop the mock and use `vi.setSystemTime(new Date('2025-08-20'))` against the static `'2025-08-14_2025-08-25'` already set in `setupTests.tsx` — that removes the leak and covers the date-only format at the same time (see my comment on `date.utility.test.ts`).

packages/pn-pa-webapp/public/locales/de/statistics.json
```diff
@@ -90,5 +90,9 @@
     "no_data_found": "Für das gewählte Zeitintervall stehen keine Daten zur Verfügung",
     "not_enough_data": "Die Anzahl der von der Körperschaft auf SEND geladenen Zustellungen reicht derzeit nicht aus, um die Versandstatistiken zu verarbeiten und anzuzeigen. Erkunde den <0>Abschnitt Zustellungen</0>, um die bisher geladenen einzusehen.",
     "notification_section_aria_label": "Abschnitt Zustellungen"
+  },
+  "maintenance_alert": {
+    "title": "Wartungsarbeiten im Gange",
+    "description": "Aufgrund einer geplanten Wartung kann es zu Verzögerungen bei der Aktualisierung der Daten im Dashboard. Bitte beachten Sie das auf dieser Seite angegebene Datum für den Stand der letzten Aktualisierung. Nach Abschluss der Wartung werden die Daten wieder regelmäßig aktualisiert."
```
fbianchicodermine: This sentence is missing its verb: `"...kann es zu Verzögerungen bei der Aktualisierung der Daten im Dashboard."` needs `kommen` at the end — `es kann zu ... kommen` is the required construction.

```suggestion
    "description": "Aufgrund einer geplanten Wartung kann es zu Verzögerungen bei der Aktualisierung der Daten im Dashboard kommen. Bitte beachten Sie das auf dieser Seite angegebene Datum für den Stand der letzten Aktualisierung. Nach Abschluss der Wartung werden die Daten wieder regelmäßig aktualisiert."
```

Worth confirming with whoever owns the translations before applying.

fbianchicodermine (changes requested):
Focused pass on the confirmed bugs — 5 inline comments. Two are test-only issues, one is a real timezone edge in `isDateInRange`, one is cross-test state leakage, one is a copy fix.

packages/pn-pa-webapp/src/services/configuration.service.ts
```diff
@@ -63,6 +85,10 @@ class PaConfigurationValidator extends Validator<PaConfiguration> {
       .isOneOf(Object.values(PhysicalAddressLookupConfig));
     this.ruleFor('ACCESSIBILITY_LINK').isString().isRequired();
     this.ruleFor('SERCQ_SERVICE_STATEMENT_LINK').isString().isRequired();
+    this.ruleFor('STATISTICS_MAINTENANCE_DATES')
+      .isString()
+      .isRequired()
+      .customValidator(statisticsMaintenanceDatesValidator);
```
fbianchicodermine, line 93: Is it intentional that this key stays required even when no maintenance is planned?

`Configuration.load` throws on a validation failure and `main.tsx` renders nothing in that case, so an environment whose `config.json` lacks the key doesn't just lose the alert — the whole app fails to boot on a blank page. That makes it a hard dependency on the environment configs: they all have to carry the key before this ships, and keep carrying it during all the periods with no maintenance planned.

If it is intentional, what value should ops set when there is no maintenance window? An empty string is rejected by `statisticsMaintenanceDatesValidator` (`parts.length !== 2`), so there is no "off" value today.

packages/pn-pa-webapp/src/pages/__test__/Statistics.page.test.tsx
```diff
@@ -312,6 +312,23 @@ describe('Statistics Page tests', () => {
...
+      preloadedState: {
+        userState: { user: userResponse },
+      },
+    });
+
+    const maintenanceAlert = await findByTestId('maintenanceAlert');
+    expect(maintenanceAlert).toBeInTheDocument();
+    expect(maintenanceAlert).toHaveTextContent('maintenance_alert.title');
+    expect(maintenanceAlert).toHaveTextContent('maintenance_alert.description');
+    vi.getRealSystemTime();
```
fbianchicodermine: `vi.getRealSystemTime()` is a getter: it returns the real timestamp and does not restore the clock, so the fake time set on line 316 leaks into every test declared after this one. Verified in isolation:

```
test A: vi.setSystemTime(new Date('2025-08-20')); ... vi.getRealSystemTime();
test B: new Date().toISOString()  ->  2025-08-20T00:00:00.000Z   // still frozen
```

With `vi.useRealTimers()` instead, test B sees the real date again.

Worth noting this leak is wider than the `statisticsUnderMaintenance` flag it replaced: it freezes `Date` globally for the rest of the file rather than just the config value, so the next test added here that does any date arithmetic would fail for a fairly non-obvious reason.

```suggestion
    vi.useRealTimers();
```

Sturdier still would be moving it to the existing `afterEach` at line 52 — `vi.useRealTimers()` is a no-op when timers aren't faked, so it is safe there unconditionally, and it would also run when the assertions above throw and the test never reaches this line.

fbianchicodermine (changes requested):
Two follow-ups on the CR round: one open question on the config key, one confirmed leak in the maintenance test. The other three points look correctly addressed — thanks.

fbianchicodermine (approved):

PR #2001, opened 2026-08-19
fix(PN-20948): update weight and color in SenderContact

SarahDonvito:
## Short description
update weight and color in SenderContact

- update weight of phone number
- update color in title

## How to test
Run PF/PG and check section is like [Figma](https://www.figma.com/design/ZFpWJMFvzFpAmCLG66nREW/Comunicazioni-Bonarie?node-id=3788-40008&t=e43hTYFP9zSY3CkW-4)

## Checklist


packages/pn-commons/src/components/PnSenderContacts/PnSenderContacts.tsx
```diff
@@ -71,7 +71,7 @@ const PnSenderContacts = ({ phone, site }: SenderContactsProps) => {
                       justifyContent="space-between"
                       sx={{ width: '100%' }}
                     >
-                      <Typography variant="body2" color="text.primary">
```
mflauti, line 74: `fontWeight={500}` does not visually match the semibold style shown in figma using the actual frontend font (`Titillium Web` instead of `Titillio`). I tested `fontWeight={600}` and the result appears much closer to the expected design. So, I'd suggest using `600` here to better match the intended visual result.

packages/pn-commons/src/components/PnSenderContacts/PnSenderContacts.tsx
```diff
@@ -71,7 +71,7 @@ const PnSenderContacts = ({ phone, site }: SenderContactsProps) => {
...
+                      <Typography variant="body2" color="text.primary" fontWeight={500}>
                         {phone}
                       </Typography>
```
mflauti, line 77: `CopyToClipboard` is currently rendered as an icon-only button with an empty `aria-label` in its default state.
The component uses `tooltipBefore` for the aria-label when `copied===false`, but in this case only the tooltip prop is provided, so the button has no accessible name before the copy action.
We could provide a meaningful `tooltipBefore` prop (e.g. `Copy phone number` or even something simpler, like `Copy`) so the button is properly exposed to screen readers.

mflauti (changes requested):

packages/pn-commons/src/components/PnSenderContacts/PnSenderContacts.tsx
SarahDonvito: We will update typography and it'll be update too, so in this way it's not needed to make a second action here. Don't you think?

packages/pn-commons/src/components/PnSenderContacts/PnSenderContacts.tsx
SarahDonvito: I'm going to open an a11y issue in mui-italia repo

packages/pn-commons/src/components/PnSenderContacts/PnSenderContacts.tsx
mflauti: Got it, thanks. Since the typography will be updated to use the actual `Titillio` font used in figma, I agree that it makes sense to keep `500` here!

mflauti (dismissed):

mflauti (approved):

PR #2003, opened 2026-08-20
fix(PN-20966): resize column in desktop notifications table

SarahDonvito:
## Short description
resize column in desktop notifications table

## How to test
Run PF and PG and check the different size screen

## Checklist


packages/pn-personafisica-webapp/src/components/Notifications/DesktopNotifications.tsx
```diff
@@ -43,32 +43,43 @@ const DesktopNotifications = ({
   const { t } = useTranslation('notifiche');
   const filterNotificationsRef = useRef({ filtersApplied: false, cleanFilters: () => void 0 });
 
+  // 10 + 8 + x + y + z = 100
```
mflauti: Could we remove these comments or update them to match the actual values?
For example, the comment says:
`xl -> iun = 20 ... subject = 36` while the implementation uses `iun = 26` an `subject = 28`.
As written, these comments are misleading and may quickly become outdated

mflauti (commented):
I verified that the change fixes the truncation of the Sender column in the reported case.
However, the same issue still occurs for the Subject column: even on very large screens the subject is truncated while other columns have plenty of unused space.
Is the scope of this task specifically limited to the Sender column or should the table generally avoid truncating fields when additional space is available?

packages/pn-personafisica-webapp/src/components/Notifications/DesktopNotifications.tsx
SarahDonvito: I'll remove it, it was a mistake

mflauti (approved):

PR #2004, opened 2026-08-20
feat(PN-20351): PF/PG - Rework timeline layout

alessandrogelmi:
## Short description
Include a summary of the changes.

- Feature A
- Feature B

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

## Checklist


packages/pn-personafisica-webapp/src/pages/__test__/NotificationTimeline.page.test.tsx
```diff
@@ -0,0 +1,215 @@
...
+        }
+      );
+    });
+
+    expect(result.getByTestId('api-error-getReceivedNotificationTimeline')).toBeInTheDocument();
+    expect(result.queryByTestId('NotificationEventsTimeline')).not.toBeInTheDocument();
+  });
+});
+
+describe('NotificationTimeline Page - new timeline disabled', () => {
```
AndreaCimini90, line 223: The download coverage is currently scoped to the `IS_NEW_TIMELINE_ENABLED = false` suite, so it only exercises the legacy response. However, this PR changes the handler to obtain the IUN and legal facts from the new timeline response as well.

Could we add a test with the flag enabled that clicks a legal-fact action rendered by `NotificationEventsTimeline` and verifies the document request uses `NotificationTimelineResponse.iun` and the expected document ID? It would also be useful to cover the AAR branch, since it uses a different document type and ID format.

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -61,12 +63,24 @@ const NotificationTimeline: React.FC = () => {
   const delegatorsFromStore = useAppSelector(
     (state: RootState) => state.generalInfoState.delegators
   );
-
   const notification = useAppSelector((state: RootState) => state.notificationState.notification);
+  const notificationTimeline = useAppSelector(
+    (state: RootState) => state.notificationState.notificationTimeline
+  );
 
   const isCancelled = useIsCancelled({ notification });
```
AndreaCimini90: When the new timeline is enabled, only `notificationTimeline` is populated, but the cancellation state is still derived from the legacy `notification` object. As a result, a cancelled notification returned by the new endpoint may not show the cancellation alert, and the download guard may remain inactive.

Please derive the cancellation state from the active data source (or normalize both responses into a shared view model) and add a test with the feature flag enabled that verifies the alert is shown and non-cancellation legal facts cannot trigger a document request.

AndreaCimini90 (changes requested):

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -61,12 +63,28 @@ const NotificationTimeline: React.FC = () => {
...
+  const notificationTimeline = useAppSelector(
+    (state: RootState) => state.notificationState.notificationTimeline
+  );
 
-  const isCancelled = useIsCancelled({ notification });
+  const isCancelled = useIsCancelled({
+    notification: IS_NEW_TIMELINE_ENABLED ? notificationTimeline : notification,
+  });
+
+  console.log({ isCancelled });
```
AndreaCimini90: remove this

AndreaCimini90 (changes requested):

AndreaCimini90 (approved):

PR #2015, opened 2026-08-26
feat(PN-21141): add MIBreadcrumbs on portals

SarahDonvito:
## Short description
Replace PnBreadcrumb with MIBreadcrumbs

- Add MIBreadcrumbs in PA
- Add MIBreadcrumbs in PF
- Add MIBreadcrumbs in PG

## How to test
Login as PF/PG/PA and check breadcrumb works fine and it is like this [Figma](https://www.figma.com/design/pvhsJpPGQbEwNLiyP7BOIw/MUI-Italia-Next---Design-System?node-id=123-7768&t=vDgZGjlyFXyrulEE-4)

## Checklist


packages/pn-pa-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -112,20 +111,18 @@ const NotificationTimeline: React.FC = () => {
...
+      <MIBreadcrumbs
+        backButtonLabel={t('button.indietro', { ns: 'common' })}
+        backButtonAction={() => navigate(routes.DASHBOARD)}
+      >
+        <MIBreadcrumbItem
+          label={t('detail.breadcrumb-root', { ns: 'notifiche' })}
+          onClick={() => navigate(routes.DASHBOARD)}
+          data-testid="breadcrumb-root-button"
+        />
+        <MIBreadcrumbItem label={notification.iun} current />
```
fbianchicodermine: With `IS_NEW_TIMELINE_ENABLED` set to `true` this label renders empty.

`getSentNotificationTimeline` writes to `state.notificationState.notificationTimeline`, while `state.notificationState.notification` is only populated by `getSentNotification` — they are two separate slices in the reducer. That is exactly what `notificationIUN` (line 53) is for. The flag is `true` in `config-dev.json`, so on this page `notification.iun` stays at its initial `''` and the current breadcrumb item shows nothing.

`label={notificationIUN}` restores the previous behaviour.

packages/pn-pa-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -112,20 +111,18 @@ const NotificationTimeline: React.FC = () => {
...
+      </MIBreadcrumbs>
     );
   }, [id, i18n.language, location.state, notificationIUN]);
```
fbianchicodermine, line 139: Why is `notificationIUN` still a dependency here? The callback no longer reads it — it renders `notification.iun` directly — so this memo is keyed on a value it never uses, and does not recompute when the value it actually renders changes. Related to the comment above: using `notificationIUN` for the label would make this dependency correct again.

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -181,17 +180,26 @@ const NotificationTimeline: React.FC = () => {
...
+      >
+        <MIBreadcrumbItem
+          label={t('menu.notifiche')}
+          onClick={() =>
+            navigate(mandateId ? routes.GET_NOTIFICHE_DELEGATO_PATH(mandateId) : routes.NOTIFICHE)
+          }
+          data-testId="breadcrumb-root-button"
+        />
+        <MIBreadcrumbItem
+          label={notification.subject ?? t('menu.fallback-notification', { ns: 'notifiche' })}
```
fbianchicodermine: Same as in PA: with `IS_NEW_TIMELINE_ENABLED` set to `true` this label renders empty. The new-timeline flow populates `state.notificationState.notificationTimeline`, not `notification`, which is exactly why `notificationSubject` exists at line 82. The flag is `true` in `config-dev.json`.

Note that the fallback cannot cover this case either: the initial state is `subject: ''`, not `undefined`, so `??` never triggers on an unfetched notification. Suggest `label={notificationSubject || t('menu.fallback-notification', ...)}`.

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -181,17 +180,26 @@ const NotificationTimeline: React.FC = () => {
...
+          current
+        />
+      </MIBreadcrumbs>
     );
   }, [id, i18n.language, notificationSubject, mandateId]);
```
fbianchicodermine: Why is `notificationSubject` still a dependency here? The callback renders `notification.subject` instead, so this memo is keyed on a value it never reads. Related to the comment on line 199: using `notificationSubject` for the label would make this dependency correct again.

packages/pn-personagiuridica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -141,17 +139,18 @@ const NotificationTimeline: React.FC = () => {
...
-        currentLocationLabel={notificationSubject ?? ''}
-        goBackAction={() => navigate(backRoute)}
-      />
+      <MIBreadcrumbs>
+        <MIBreadcrumbItem
+          label={t('menu.notifiche', { ns: 'notifiche' })}
+          href={mandateId ? routes.NOTIFICHE_DELEGATO : routes.NOTIFICHE}
+        />
+        <MIBreadcrumbItem
+          label={notification.subject ?? t('menu.fallback-notification', { ns: 'notifiche' })}
```
fbianchicodermine: Same as PA and PF: with `IS_NEW_TIMELINE_ENABLED` set to `true` this renders empty, because the new-timeline flow fills `notificationTimeline` and not `notification` — hence `notificationSubject` at line 69. The flag is `true` in `config-dev.json`.

Same caveat on the fallback: the initial `subject` is `''`, not `undefined`, so `??` never kicks in — it would need `||`.

packages/pn-personagiuridica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -141,17 +139,18 @@ const NotificationTimeline: React.FC = () => {
...
+          current
+        />
+      </MIBreadcrumbs>
     );
   }, [id, i18n.language, notificationSubject, mandateId]);
```
fbianchicodermine: Same question as in PF: `notificationSubject` is still listed as a dependency, but the callback renders `notification.subject` instead — the memo is keyed on a value it never reads. Using `notificationSubject` for the label would make this dependency correct again.

fbianchicodermine (changes requested):
Change request on the breadcrumb current label of the three `NotificationTimeline` pages: it renders empty when `IS_NEW_TIMELINE_ENABLED` is `true`. Details inline.

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
fbianchicodermine: Screenshot 2026-09-02 alle 08 55 37

packages/pn-personagiuridica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -438,13 +437,20 @@ const NotificationDetail = () => {
...
-      <PnBreadcrumb
-        showBackAction={!fromQrCode}
-        linkRoute={backRoute}
-        linkLabel={breadcrumbLabel}
-        currentLocationLabel={notification.subject ?? ''}
-        goBackAction={() => navigate(backRoute)}
-      />
+      <MIBreadcrumbs
+        backButtonLabel={t('button.indietro', { ns: 'common' })}
+        backButtonAction={() => navigate(routes.DELEGATI)}
```
fbianchicodermine: `routes.DELEGATI` is `/deleghe/delegati`, so on mobile the back button takes the user from a notification detail to the delegations list. `backRoute` (line 433) already holds the right destination and is used by the root item on line 446 — this should be `navigate(backRoute)`, which is also what the previous `goBackAction` did.

packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -436,15 +435,24 @@ const NotificationDetail: React.FC = () => {
...
-        linkLabel={
-          mandateId || delegatorsFromStore.length > 0 ? breadcrumbLabel : t('menu.notifiche')
+      <MIBreadcrumbs
+        backButtonLabel={
+          rapidAccessSource ? t('menu.notifiche') : t('button.indietro', { ns: 'common' })
         }
-        currentLocationLabel={notification.subject ?? ''}
-        goBackAction={() => navigate(backRoute)}
-      />
+        backButtonAction={() => navigate(routes.NOTIFICHE)}
```
fbianchicodermine: This ignores `backRoute` (line 427), which the root item on line 448 does use: for a delegate viewing a delegator's notification (`mandateId` set), the mobile back button lands on the user's own inbox instead of the delegator's. The previous `goBackAction` was `navigate(backRoute)` — same fix here.

packages/pn-pa-webapp/src/pages/NewApiKey.page.tsx
```diff
@@ -93,12 +88,17 @@ const NewApiKey = () => {
...
               <Grid item xs={12} lg={8}>
-                <PnBreadcrumb
-                  linkRoute={routes.API_KEYS}
-                  linkLabel={t('title')}
-                  currentLocationLabel={tkp('page-title')}
-                  goBackLabel={t('button.indietro', { ns: 'common' })}
-                />
+                <MIBreadcrumbs
+                  backButtonLabel={t('button.indietro', { ns: 'common' })}
+                  backButtonAction={() => navigate(routes.DASHBOARD)}
```
fbianchicodermine: The parent of this page is the API keys list, not the dashboard: the root item on line 97 navigates to `routes.API_KEYS`, and the back button should match. As written, leaving the new-API-key form drops the user on the notifications dashboard.

packages/pn-pa-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -112,20 +111,18 @@ const NotificationTimeline: React.FC = () => {
...
-        goBackAction={() =>
-          location.state?.fromNotificationDetail
-            ? navigate(-1)
-            : navigate(backRoute, { replace: true })
-        }
-        goBackLabel={t('button.indietro', { ns: 'common' })}
-      />
+      <MIBreadcrumbs
+        backButtonLabel={t('button.indietro', { ns: 'common' })}
+        backButtonAction={() => navigate(routes.DASHBOARD)}
```
fbianchicodermine: This drops the previous back logic:

```js
location.state?.fromNotificationDetail
  ? navigate(-1)
  : navigate(routes.GET_DETTAGLIO_NOTIFICA_PATH(id), { replace: true })
```

The level immediately above the timeline is the notification detail, not the dashboard, so on mobile the user now skips it entirely. Note that `location.state` is still listed in the `useMemo` deps even though nothing reads it any more.

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -181,17 +180,26 @@ const NotificationTimeline: React.FC = () => {
...
     return (
-      <PnBreadcrumb
-        linkRoute={mandateId ? routes.GET_NOTIFICHE_DELEGATO_PATH(mandateId) : routes.NOTIFICHE}
-        linkLabel={t('menu.notifiche')}
-        currentLocationLabel={notificationSubject ?? ''}
-        goBackAction={() => navigate(backRoute)}
-      />
+      <MIBreadcrumbs
+        backButtonLabel={t('button.indietro', { ns: 'common' })}
+        backButtonAction={() =>
```
fbianchicodermine: Same as PA: the previous `goBackAction` navigated to the notification detail (`GET_DETTAGLIO_NOTIFICA_DELEGATO_PATH` / `GET_DETTAGLIO_NOTIFICA_PATH`), which is the level immediately above the timeline. Now the back button and the root item (line 193) both go to the notifications list, so the detail page is no longer reachable from the breadcrumb on mobile.

fbianchicodermine (changes requested):
Change request on the `backButtonAction` destinations.

Worth stating up front, because it makes these easy to miss: in `@pagopa/mui-italia@2.7.0-RC.0`, `MIBreadcrumbs` renders **either** the back button **or** the children, never both — below the `sm` breakpoint (or with `variant="compact"`) it renders only `<MIBreadcrumbItem type="back" onClick={backButtonAction} />` and drops the children; from `sm` up it renders only the children and no back button at all. So `backButtonAction` is the only breadcrumb navigation available on mobile. There is no `matchMedia` mock in the test setup, so `useMediaQuery` is always `false` in jsdom and none of this is covered by the updated tests.

The Figma rules state "Mobile → la breadcrumb è una e porta al livello di flusso immediatamente precedente rispetto alla pagina corrente", but in five places the back button points somewhere else — and in most of them the correct route is already computed a few lines above and used by the root item.

packages/pn-personagiuridica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -141,17 +139,18 @@ const NotificationTimeline: React.FC = () => {
...
     return (
-      <PnBreadcrumb
-        linkRoute={mandateId ? routes.NOTIFICHE_DELEGATO : routes.NOTIFICHE}
-        linkLabel={t('menu.notifiche')}
-        currentLocationLabel={notificationSubject ?? ''}
-        goBackAction={() => navigate(backRoute)}
-      />
+      <MIBreadcrumbs>
+        <MIBreadcrumbItem
+          label={t('menu.notifiche', { ns: 'notifiche' })}
```
fbianchicodermine: This one is always rendered, so the breadcrumb root shows the literal string `menu.notifiche` in every language. `menu.notifiche` lives in `common.json` (`"In arrivo"` for PG), not in `notifiche.json`, which has no `menu` section at all. The previous code read it without an explicit `ns` and resolved against the default namespace. Dropping `{ ns: 'notifiche' }` fixes it.

packages/pn-personagiuridica-webapp/src/pages/NotificationTimeline.page.tsx
fbianchicodermine: Same `ns` problem as line 146, for `menu.fallback-notification`: the key is in `common.json`. (See also my earlier comment on this line about `notificationSubject` and `??` vs `||`.)

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
fbianchicodermine: Same `ns` problem: `menu.fallback-notification` was added to `common.json`, not to `notifiche.json`, so this lookup renders the raw key. Drop `{ ns: 'notifiche' }`. (See also my earlier comment on this line.)

packages/pn-personafisica-webapp/src/pages/InformalNotificationDetail.page.tsx
```diff
@@ -193,17 +192,24 @@ const InformalNotificationDetail: React.FC = () => {
...
+          onClick={() => navigate(routes.NOTIFICHE)}
+          label={
+            delegatorsFromStore.length > 0
+              ? t('menu.notifiche-utente', { ns: 'common' })
+              : t('menu.notifiche')
+          }
+          data-testId="breadcrumb-root-button"
+        />
+        <MIBreadcrumbItem
+          label={primaryMessage?.subject ?? t('menu.fallback-communication', { ns: 'notifiche' })}
```
fbianchicodermine: Here the fallback really does fire, since `primaryMessage?.subject` can be `undefined` — so this is user-visible: the breadcrumb shows the literal string `menu.fallback-communication`. The key is in `common.json` under `menu`, and this page's default namespace is already `common`, so `t('menu.fallback-communication')` is enough.

packages/pn-personagiuridica-webapp/src/pages/InformalNotificationDetail.page.tsx
```diff
@@ -157,19 +155,22 @@ const InformalNotificationDetail: React.FC = () => {
...
+            currentRecipient?.denomination
+              ? t('menu.notifiche-impresa', {
+                  organization: currentOrganizationName ?? '',
+                })
+              : t('menu.notifiche')
+          }
+          href={routes.NOTIFICHE}
+        />
+        <MIBreadcrumbItem
+          label={primaryMessage?.subject ?? t('menu.fallback-communication', { ns: 'notifiche' })}
```
fbianchicodermine: Same as PF: `primaryMessage?.subject` can be `undefined`, so the fallback fires and the breadcrumb shows the literal `menu.fallback-communication`. Drop `{ ns: 'notifiche' }`.

packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -436,15 +435,24 @@ const NotificationDetail: React.FC = () => {
...
+      >
+        <MIBreadcrumbItem
+          label={
+            mandateId || delegatorsFromStore.length > 0 ? breadcrumbLabel : t('menu.notifiche')
+          }
+          onClick={() => navigate(backRoute)}
+          data-testId="breadcrumb-root-button"
+        />
+        <MIBreadcrumbItem
+          label={notification.subject ?? t('menu.fallback-notification', { ns: 'notifiche' })}
```
fbianchicodermine: Same wrong `ns` — the key is in `common.json`. Worth noting that here the fallback never fires anyway: the initial state is `subject: ''`, not `undefined`, so `??` does not trigger. If the intent is to show a fallback while the notification is still loading, this needs `||`.

packages/pn-personagiuridica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -438,13 +437,20 @@ const NotificationDetail = () => {
...
+      >
+        <MIBreadcrumbItem
+          label={breadcrumbLabel}
+          onClick={() => navigate(backRoute)}
+          data-testid="breadcrumb-root-button"
+        />
+        <MIBreadcrumbItem
+          label={notification.subject ?? t('menu.fallback-notification', { ns: 'notifiche' })}
```
fbianchicodermine: Same as PF: wrong `ns` (the key is in `common.json`), and the `??` never triggers because the initial `subject` is `''` rather than `undefined` — `||` would be needed for the fallback to have any effect.

fbianchicodermine (changes requested):
Change request on the i18n of the new breadcrumb labels: the `menu.fallback-*` keys were added to `common.json`, but they are read with `{ ns: 'notifiche' }`. `notifiche.json` has no `menu` section, and no `fallbackNS` is configured in any `i18n.ts` (only `fallbackLng`), so those lookups fail and i18next renders the raw key. All these pages already default to the `common` namespace via `useTranslation(['common', 'notifiche', ...])`, so dropping `{ ns: 'notifiche' }` is enough.

packages/pn-personafisica-webapp/src/pages/InformalNotificationDetail.page.tsx
```diff
@@ -193,17 +192,24 @@ const InformalNotificationDetail: React.FC = () => {
...
+        backButtonAction={() => navigate(routes.NOTIFICHE)}
+      >
+        <MIBreadcrumbItem
+          onClick={() => navigate(routes.NOTIFICHE)}
+          label={
+            delegatorsFromStore.length > 0
+              ? t('menu.notifiche-utente', { ns: 'common' })
+              : t('menu.notifiche')
+          }
+          data-testId="breadcrumb-root-button"
```
fbianchicodermine: `React does not recognize the `data-testId` prop on a DOM element`

use `data-testid`

packages/pn-personafisica-webapp/src/pages/NotificationDetail.page.tsx
```diff
@@ -436,15 +435,24 @@ const NotificationDetail: React.FC = () => {
...
-        goBackAction={() => navigate(backRoute)}
-      />
+        backButtonAction={() => navigate(routes.NOTIFICHE)}
+      >
+        <MIBreadcrumbItem
+          label={
+            mandateId || delegatorsFromStore.length > 0 ? breadcrumbLabel : t('menu.notifiche')
+          }
+          onClick={() => navigate(backRoute)}
+          data-testId="breadcrumb-root-button"
```
fbianchicodermine: `React does not recognize the `data-testId` prop on a DOM element`

use `data-testid`

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -181,17 +180,26 @@ const NotificationTimeline: React.FC = () => {
...
+          navigate(mandateId ? routes.GET_NOTIFICHE_DELEGATO_PATH(mandateId) : routes.NOTIFICHE)
+        }
+      >
+        <MIBreadcrumbItem
+          label={t('menu.notifiche')}
+          onClick={() =>
+            navigate(mandateId ? routes.GET_NOTIFICHE_DELEGATO_PATH(mandateId) : routes.NOTIFICHE)
+          }
+          data-testId="breadcrumb-root-button"
```
fbianchicodermine: `React does not recognize the `data-testId` prop on a DOM element`

use `data-testid`

packages/pn-personafisica-webapp/src/pages/NuovaDelega.page.tsx
```diff
@@ -222,24 +226,19 @@ const NuovaDelega = () => {
...
-        currentLocationLabel={t('nuovaDelega.breadcrumb')}
-      />
+      <MIBreadcrumbs
+        backButtonLabel={t('button.indietro', { ns: 'common' })}
+        backButtonAction={() => navigate(routes.DELEGHE)}
+      >
+        <MIBreadcrumbItem
+          label={t('nuovaDelega.title')}
+          onClick={() => navigate(routes.DELEGHE)}
+          data-testId="breadcrumb-root-button"
```
fbianchicodermine: `React does not recognize the `data-testId` prop on a DOM element`

use `data-testid`

fbianchicodermine (changes requested):

packages/pn-personagiuridica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -141,17 +139,15 @@ const NotificationTimeline: React.FC = () => {
...
-      : routes.GET_DETTAGLIO_NOTIFICA_PATH(id);
 
     return (
-      <PnBreadcrumb
-        linkRoute={mandateId ? routes.NOTIFICHE_DELEGATO : routes.NOTIFICHE}
-        linkLabel={t('menu.notifiche')}
-        currentLocationLabel={notificationSubject ?? ''}
-        goBackAction={() => navigate(backRoute)}
-      />
+      <MIBreadcrumbs>
```
fbianchicodermine: No `backButtonLabel` / `backButtonAction` here, so on mobile this renders the `mui-italia` defaults:

- `backButtonLabel = "Indietro"`, hardcoded in the component, so it stays Italian in the de/en/fr/sl bundles;
- `backButtonAction = () => window.history.back()`, which is not the same as navigating to the level above (the notification detail): it replays whatever the user did before, and does nothing at all when the page was opened directly from a link.

Every other page in this PR passes `backButtonLabel={t('button.indietro', { ns: 'common' })}` plus an explicit action — this one should too. Note that `const navigate = useNavigate()` was removed from this file, so it needs to come back.

packages/pn-personagiuridica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -141,17 +139,15 @@ const NotificationTimeline: React.FC = () => {
...
+        <MIBreadcrumbItem
+          label={t('menu.notifiche')}
+          href={mandateId ? routes.NOTIFICHE_DELEGATO : routes.NOTIFICHE}
```
fbianchicodermine: `href` makes `MIBreadcrumbItem` render a plain `<a>`, so this is a full page reload: the SPA is re-bootstrapped, the Redux store is thrown away, and both the react-router navigation and any `Prompt` blocker are bypassed. Every other page in this PR uses `onClick={() => navigate(...)}`.

packages/pn-personagiuridica-webapp/src/pages/InformalNotificationDetail.page.tsx
```diff
@@ -157,19 +155,22 @@ const InformalNotificationDetail: React.FC = () => {
...
-          currentRecipient?.denomination
-            ? t('menu.notifiche-impresa', {
-                organization: currentOrganizationName ?? '',
-              })
-            : t('menu.notifiche')
-        }
-        currentLocationLabel={primaryMessage?.subject ?? ''}
-        goBackAction={() => navigate(routes.NOTIFICHE)}
-      />
+      <MIBreadcrumbs>
```
fbianchicodermine: Same as `NotificationTimeline`: with no `backButtonLabel` / `backButtonAction`, the mobile back button falls back to the hardcoded `"Indietro"` (untranslated outside Italian) and to `window.history.back()` instead of navigating to the notifications list. `useNavigate` was removed from this file as well and would need to come back.

packages/pn-personagiuridica-webapp/src/pages/InformalNotificationDetail.page.tsx
```diff
@@ -157,19 +155,22 @@ const InformalNotificationDetail: React.FC = () => {
...
+        <MIBreadcrumbItem
+          label={
+            currentRecipient?.denomination
+              ? t('menu.notifiche-impresa', {
+                  organization: currentOrganizationName ?? '',
+                })
+              : t('menu.notifiche')
+          }
+          href={routes.NOTIFICHE}
```
fbianchicodermine: Same as `NotificationTimeline`: `href` renders a plain `<a>` and causes a full page reload instead of a client-side navigation — should be `onClick={() => navigate(routes.NOTIFICHE)}`.

fbianchicodermine (changes requested):
Change request on the two PG pages that use `<MIBreadcrumbs>` without the back props and with `href` on the parent item — they behave differently from every other page in this PR.

As noted in my earlier review, `MIBreadcrumbs` renders only the back button below the `sm` breakpoint and drops the children entirely, so on mobile these two pages fall back to the component defaults: the hardcoded Italian string `"Indietro"` and `window.history.back()`.

packages/pn-personagiuridica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -141,17 +142,36 @@ const NotificationTimeline: React.FC = () => {
...
+      >
+        <MIBreadcrumbItem
+          label={breadcrumbLabel}
+          onClick={() => {
+            navigate(mandateId ? routes.NOTIFICHE_DELEGATO : routes.NOTIFICHE);
+          }}
+          data-testid="breadcrumb-root-button"
+        />
+        <MIBreadcrumbItem
+          label={notification.subject || t('menu.fallback-notification')}
```
fbianchicodermine: This brings back the issue from my first review: it reads `notification.subject` instead of `notificationSubject` (line 70). With `IS_NEW_TIMELINE_ENABLED` set to `true` — the value in every `config-dev.json` — the data lives in `notificationTimeline`, so this level falls through to the generic `t('menu.fallback-notification')` ("Dettaglio notifica") instead of showing the actual subject. PA (line 129) and PF (line 198) correctly use the derived values.

It also makes `notificationSubject` a dead dependency again on line 176.

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -181,17 +180,29 @@ const NotificationTimeline: React.FC = () => {
...
-        currentLocationLabel={notificationSubject ?? ''}
-        goBackAction={() => navigate(backRoute)}
-      />
+      <MIBreadcrumbs
+        backButtonLabel={t('button.indietro', { ns: 'common' })}
+        backButtonAction={() => navigate(notificationDetailRoute)}
+      >
+        <MIBreadcrumbItem
+          label={t('menu.notifiche')}
+          onClick={() => navigate(routes.NOTIFICHE)}
```
fbianchicodermine: The `mandateId` branch was dropped here — this used to be `navigate(mandateId ? routes.GET_NOTIFICHE_DELEGATO_PATH(mandateId) : routes.NOTIFICHE)`, and on `develop` the `PnBreadcrumb` `linkRoute` had the same branch. So a delegate looking at the timeline of a delegator's notification now lands on their own inbox when clicking the first level. PG kept the branch (line 161) and `NotificationDetail` uses it too, so PF is the only one out of line.

packages/pn-personagiuridica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -141,17 +142,36 @@ const NotificationTimeline: React.FC = () => {
...
+          onClick={() => navigate(notificationDetailRoute)}
+        />
+        <MIBreadcrumbItem
+          label={t('detail.notification-timeline-section.title', { ns: 'notifiche' })}
+          current
+        />
+      </MIBreadcrumbs>
     );
   }, [id, i18n.language, notificationSubject, mandateId]);
```
fbianchicodermine: `organization?.name` is now read by `breadcrumbLabel` (line 152) but is not in this dependency array — `NotificationDetail` lists it (`[fromQrCode, i18n.language, notification.subject, mandateId, organization?.name]`). If the organization is not in the store yet on the first render, the label stays as "In arrivo per " with no name.

packages/pn-pa-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -112,20 +111,29 @@ const NotificationTimeline: React.FC = () => {
...
         }
-        goBackLabel={t('button.indietro', { ns: 'common' })}
-      />
+      >
+        <MIBreadcrumbItem
+          label={t('detail.breadcrumb-root', { ns: 'notifiche' })}
+          onClick={() => navigate(routes.DASHBOARD)}
+          data-testid="breadcrumb-root-button"
+        />
+        <MIBreadcrumbItem
```
fbianchicodermine, line 128: Minor: the new middle level is clickable but has no `data-testid`, unlike the root item on line 126 — same in PF (line 197) and PG (line 166). Worth adding, especially since none of the three `NotificationTimeline.page.test.tsx` files asserts anything on the breadcrumb, so this whole new level is currently uncovered.

fbianchicodermine (changes requested):
The extra breadcrumb level looks right — root → notification → "Stato della notifica" matches the desktop pattern in Figma, the new `detail.notification-timeline-section.title` key is present in all three packages for all five languages, and on mobile the back button still points at the notification detail in all three portals. Two regressions came in with it, though.

packages/pn-personafisica-webapp/src/pages/NuovaDelega.page.tsx
```diff
@@ -222,24 +226,19 @@ const NuovaDelega = () => {
...
-          </Fragment>
-        }
-        currentLocationLabel={t('nuovaDelega.breadcrumb')}
-      />
+      <MIBreadcrumbs
+        backButtonLabel={t('button.indietro', { ns: 'common' })}
+        backButtonAction={() => navigate(routes.DELEGHE)}
+      >
+        <MIBreadcrumbItem
+          label={t('nuovaDelega.title')}
```
fbianchicodermine: Parent should be "Deleghe"

Image

fbianchicodermine (changes requested):

packages/pn-personafisica-webapp/src/pages/NotificationTimeline.page.tsx
```diff
@@ -181,17 +180,44 @@ const NotificationTimeline: React.FC = () => {
...
+          onClick={() => navigate(notificationDetailRoute)}
+          data-testid="breadcrumb-subject-button"
+        />
+        <MIBreadcrumbItem
+          label={t('detail.notification-timeline-section.title', { ns: 'notifiche' })}
+          current
+        />
+      </MIBreadcrumbs>
     );
   }, [id, i18n.language, notificationSubject, mandateId]);
```
fbianchicodermine: `delegatorsFromStore` is now read inside this memo — at line 191 for `delegatorName` and at line 206 for the `delegatorsFromStore.length > 0` guard — but it is not in the dependency array. `NotificationDetail` lists it (line 457: `[rapidAccessSource, i18n.language, notification.subject, delegatorsFromStore, mandateId]`).

Same class as the `organization?.name` dependency just fixed in PG. The delegators come from an async call, so if they land after the first render the label stays at "In arrivo" instead of "In arrivo per te" or the delegator's name. A change in `notificationSubject` may re-run the memo and mask it, but that ordering is not guaranteed.

packages/pn-personafisica-webapp/src/pages/__test__/NotificationTimeline.page.test.tsx
```diff
@@ -218,6 +225,68 @@ describe('NotificationTimeline Page - IS_NEW_TIMELINE_ENABLED enabled', () => {
...
+        route: routes.GET_DETTAGLIO_NOTIFICA_TIMELINE_PATH(timelineIun),
+        path: routes.DETTAGLIO_NOTIFICA_TIMELINE,
+      });
+    });
+
+    const rootButton = result.getByTestId('breadcrumb-root-button');
+    fireEvent.click(rootButton);
+
+    await waitFor(() => {
+      expect(mockNavigateFn).toHaveBeenCalled();
```
fbianchicodermine: This asserts only that `navigate` was called, not where — so the test passes for any destination, including the wrong ones this PR went through (the root item losing the `mandateId` branch would not have been caught here). Worth asserting the exact route, as the sibling test does for the subject item. Same at PF:456, PG:292, PG:455 and PA:429.

Related: no test sets `mandateId`, so the delegate paths are untested — and those are exactly the ones where the root and back destinations differ.

packages/pn-pa-webapp/src/pages/__test__/NotificationTimeline.page.test.tsx
```diff
@@ -218,6 +224,49 @@ describe('NotificationTimeline Page - IS_NEW_TIMELINE_ENABLED enabled', () => {
...
+        route: `/${timelineIun}/dettaglio/timeline`,
+        path: '/:id/dettaglio/timeline',
+      });
+    });
+
+    const subjectButton = result.getByTestId('breadcrumb-iun-button');
+    fireEvent.click(subjectButton);
+
+    await waitFor(() => {
+      expect(mockNavigateFn).toHaveBeenCalledWith(`/dashboard/${timelineIun}/dettaglio`);
```
fbianchicodermine: The expected path is hardcoded here and at line 411 (`/dashboard/${iun}/dettaglio`), while PF and PG use `routes.GET_DETTAGLIO_NOTIFICA_PATH(...)`. Using the constant keeps the test honest if the route is ever restructured.

packages/pn-personafisica-webapp/src/pages/__test__/NotificationTimeline.page.test.tsx
```diff
@@ -37,8 +37,15 @@ vi.mock('../../services/configuration.service', async () => {
   };
 });
 
+const mockNavigateFn = vi.fn();
+vi.mock('react-router-dom', async () => ({
```
fbianchicodermine: Mocking `useNavigate` for the whole file diverges from the convention used elsewhere in the repo — `NotificationDetail.page.test.tsx` asserts on `result.router.state.location.pathname`, which checks that the navigation actually resolves to a route rather than just checking the argument passed to a spy. Not blocking, but the router-state assertion is the stronger one and keeps these files consistent with their neighbours.

fbianchicodermine (dismissed):

fbianchicodermine (approved):

PR #2020, opened 2026-08-28
fix(PN-20892): move support route out of the ToS guard

fbianchicodermine:
## Short description

On the ToS acceptance page, clicking the *Assistenza* button in the header did nothing and the button disappeared right after the click.

The `/assistenza` route was declared inside `ToSGuard`, so navigating to it rendered `ToSAcceptancePage` again, while the layout already hides the assistance button on that path. The support page itself needs no user consent: it reads no user state, and the email is entered manually in the form.

- Moved the `/assistenza` route out of `ToSGuard` (and out of `RapidAccessGuard`/`OnboardingGuard`), keeping it under `SessionGuard` so the session token is still attached to the Zendesk authorization call
- Added `navigation/__test__/routes.test.tsx`, covering the support route with ToS and privacy not accepted, plus the counter-check that a guarded route still shows the ToS page

## How to test

Log in with a user that has not accepted the ToS, or mock `https://webapi.dev.notifichedigitali.it/bff/v2/tos-privacy?type=TOS%2CDATAPRIVACY` with Netify, returning an empty array `[]` in the response body.

Then, on the ToS acceptance page:

1. Click *Assistenza* in the header: the support page opens instead of the ToS page reloading
2. Submit the form with a valid email: the redirect to Zendesk works
3. Go back: the ToS acceptance page is shown again

## Checklist

### DIFF
--- a/packages/pn-personafisica-webapp/src/navigation/__test__/routes.test.tsx
+++ b/packages/pn-personafisica-webapp/src/navigation/__test__/routes.test.tsx
@@ -1,0 +1,70 @@
+import MockAdapter from 'axios-mock-adapter';
+
+import { userResponse } from '../../__mocks__/Auth.mock';
+import { tosPrivacyConsentMock } from '../../__mocks__/Consents.mock';
+import { act, render, screen, waitFor } from '../../__test__/test-utils';
+import { apiClient } from '../../api/apiClients';
+import Router from '../routes';
+import * as routes from '../routes.const';
+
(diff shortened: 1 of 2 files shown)


SarahDonvito (commented):
One thing I’d consider is adding a test that covers the actual user flow from the Assistance button on the ToS acceptance page, not just the route-level behavior. That would make the fix more resilient and better protect against regressions in the header/navigation interaction.

SarahDonvito (approved):

PR #2025, opened 2026-09-01
fix(PN-20762): reset pagination when switching to delegated notifications

mflauti:
## Short description
Fixes an issue where pagination state from the organization's notification list was incorrectly reused when switching to delegated notifications, causing unavailable pages to be displayed.

- reset pagination when switching between onw and delagated notifications
- add regression coverage for pagination reset

## How to test
1. Choose a delegator with few notifications in the last 120 days and create a mandate for a PG
2. Open the organization's notification list and navigate to a page after the first one
3. Switch to "Incoming from delegations" and verify the pagination is recalculated for the delegated list an no stale pages are shown
4. Switch back to organization's list and verify the pagination is correctly reset

## Checklist

### DIFF
--- a/packages/pn-personagiuridica-webapp/src/pages/Notifiche.page.tsx
+++ b/packages/pn-personagiuridica-webapp/src/pages/Notifiche.page.tsx
@@ -27,7 +27,12 @@ import { PNRole } from '../models/User';
 import { ContactSource } from '../models/contacts';
 import { contactsSelectors } from '../redux/contact/reducers';
 import { DASHBOARD_ACTIONS, getReceivedNotifications } from '../redux/dashboard/actions';
-import { setNotificationFilters, setPagination, setSorting } from '../redux/dashboard/reducers';
+import {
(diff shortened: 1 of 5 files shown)


packages/pn-personagiuridica-webapp/src/pages/Notifiche.page.tsx
```diff
@@ -196,13 +205,17 @@ const Notifiche = ({ isDelegatedPage = false }: Props) => {
   }, []);
 
   useEffect(() => {
+    if (storedIsDelegatedPage !== isDelegatedPage) {
+      dispatch(setIsDelegatedPage(isDelegatedPage));
```
AndreaCimini90, line 209: I think that we have to restore also other filters and not only the pagination. Try to set a filter in the notification list page and after navigate to delegated page; filters are maintained between the two pages.

AndreaCimini90 (changes requested):

PR #2033, opened 2026-09-07
fix(PN-21163): PA hide notification filters on empty state

gaetano-amato:
## Short description
Hide notification filters when there are no notifications and no filters are applied.
Filters remain visible when an active filter returns no results, allowing the user to modify or clear the current filters.
Notification filters are also hidden in case of API errors.

- Added `showFilters` handling to PA notification filters
- Hidden filters when the notifications list is empty
- Kept filters visible when filters are active but return no results

## How to test
1. Open the PA Notifications page.
2. Verify that filters are displayed when notifications are available.
3. Verify that filters are hidden when there are no notifications and no active filters.
4. Apply a filter that returns no results and verify that the filters remain visible.
5. Verify that pagination is displayed only when notifications are available.

### DIFF
--- a/packages/pn-pa-webapp/src/pages/__test__/Dashboard.page.test.tsx
+++ b/packages/pn-pa-webapp/src/pages/__test__/Dashboard.page.test.tsx
@@ -228,6 +228,35 @@ describe('Dashboard Page', () => {
     expect(rows[0]).toHaveTextContent(notificationsDTO.resultsPage[1].iun);
   });
 
+  it('keeps filters visible when active filter returns no notifications', async () => {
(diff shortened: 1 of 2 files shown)


packages/pn-pa-webapp/src/pages/__test__/Dashboard.page.test.tsx
SarahDonvito, line 1: I'd add a dedicated test for the case where a filter is active but the API returns no notifications. This filter should remain visible so the user can adjust or clear the current filter. The current tests cover empty state and API error cases, but not this scenario.

SarahDonvito (changes requested):

SarahDonvito (approved):

PR #2049, opened 2026-09-14
feat(PN-20236): add campaigns list to PA

mflauti:
## Short description
Add the campaigns list page for Informal Notifications

- Add the new "Comunica con SEND" / "Campagne" navigation entry
- Add campaigns list retrieval, pagination and rendering using `MITableList`
- Add redux state for campaigns
- Add tests

## How to test
- Open the PA application
- If collapsed, expand "Comunica con SEND" from the side menu and click on "Campagne"
- Verify that the campaigns list is displayed correctly
- Verify the pagination behavior

## Checklist


packages/pn-pa-webapp/public/locales/it/campaigns.json
```diff
@@ -0,0 +1,12 @@
+{
```
AndreaCimini90, line 1: The new campaigns namespace and the common menu/button keys were added only for Italian. Users selecting English, French, German, or Slovenian will receive missing-resource requests and Italian fallback text, resulting in a mixed-language UI. Please add the corresponding campaigns.json files and common.json keys for every supported locale.

packages/pn-pa-webapp/src/pages/__test__/Campaigns.page.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  afterEach(() => {
+    mock.reset();
+  });
+
+  afterAll(() => {
+    mock.restore();
+  });
+
+  it('renders campaigns and performs the initial API request', async () => {
```
AndreaCimini90, line 45: Pagination is a core behavior introduced by this PR, but the tests cover only the initial request. Please add a test that changes page and verifies that the correct nextPagesKey is sent, plus a page-size change test that verifies pagination keys are reset. This would also catch regressions in the cursor-based pagination state.

packages/pn-pa-webapp/src/pages/Campaigns.page.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const pagesToShow: Array<number> = calculatePages(
+    pagination.size,
+    totalElements,
+    Math.min(pagination.nextPagesKey.length + 1, 3),
+    pagination.page + 1
+  );
+
+  const handleChangePage = (paginationData: PaginationData) => {
+    dispatch(setPagination({ size: paginationData.size, page: paginationData.page }));
```
AndreaCimini90: We are using the same approach used for the legal notification. Over time, this approach has proven complex and difficult to maintain.

We can try new approach!

Pagination state and campaign results are currently updated through two separate Redux transitions: first setPagination, then getCampaigns.fulfilled. This makes the request flow harder to follow and may become more complex when filters and sorting are added. Could we dispatch the requested query parameters directly to getCampaigns and update pagination, applied filters, and results atomically on fulfillment? This would keep the displayed query state consistent with the returned data and avoid an intermediate Redux update.

AndreaCimini90 (changes requested):

packages/pn-pa-webapp/src/pages/Campaigns.page.tsx
mflauti: Ok, makes sense. We can dispatch `getCampaigns `directly with the requested page/size and update the pagination state on fulfillment together with the results.
When the page size changes, we'll also reset the stored `nextPagesKey` and start again from page 0 without a key, since the exitsing keys belong to the previous page size.

AndreaCimini90 (approved):

PR #2053, opened 2026-09-15
feature(PN-19913): actions for Lokalise

fbianchicodermine:
## Short description
actions workflow for lokalise and some fix on uncorrect key

### DIFF
--- a/.github/workflows/pull-lokalise.yml
+++ b/.github/workflows/pull-lokalise.yml
@@ -26,7 +26,7 @@ permissions:
 jobs:
   pull-locale:
     runs-on: ubuntu-latest
-    # TODO: RIPRISTINARE PRIMA DEL MERGE - usa i secret di repository durante il test.
+    # TODO: RIPRISTINARE PRIMA DEL MERGE - usa secret e variabili di repository durante il test.
     #       Riallineare a io-app con "environment: lokalise" quando l'environment sara' disponibile.
     # environment: lokalise
     strategy:
@@ -37,7 +37,7 @@ jobs:
       matrix:
         app: [PFL, PF, PA, PG]
     env:
-      # I codici app sono anche il suffisso dei secret LOKALISE_PROJECT_ID_<CODICE>.
+      # I codici app sono anche il suffisso delle variabili LOKALISE_PROJECT_ID_<CODICE>.
       PACKAGES: >-
         {
           "PFL": "pn-personafisica-login",
@@ -56,7 +56,7 @@ jobs:
         uses: lokalise/lokalise-pull-action@95afae3fac9dd28db1ff4887a63b1f76cbee714d # v5.3.0
         with:
           api_token: ${{ secrets.LOKALISE_API_TOKEN }}
-          project_id: ${{ secrets[format('LOKALISE_PROJECT_ID_{0}', matrix.app)] }}
+          project_id: ${{ vars[format('LOKALISE_PROJECT_ID_{0}', matrix.app)] }}
           base_lang: it
           translations_path: packages/${{ fromJson(env.PACKAGES)[matrix.app] }}/public/locales
           file_format: json
--- a/.github/workflows/pull-lokalise.yml
+++ b/.github/workflows/pull-lokalise.yml
@@ -1,4 +1,4 @@
-name: Pull locales from Lokalise.
+name: Pull locales from Lokalise
 
 on:
   workflow_dispatch:
--- a/.github/workflows/pull-lokalise.yml
+++ b/.github/workflows/pull-lokalise.yml
@@ -1,0 +1,93 @@
+name: Pull locales from Lokalise
+
+on:
+  workflow_dispatch:
+    inputs:
+      app:
+        description: 'App da sincronizzare (all = tutte e quattro)'
+        type: choice
+        required: true
+        default: all
+        options:
+          - all
+          - PFL
+          - PF
+          - PA
+          - PG
+
+concurrency:
+  group: lokalise-pull-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+  pull-requests: write
+  # Le label sulla PR passano dall'API issues, non da quella pull-requests.
+  issues: write
+
+jobs:
+  pull-locale:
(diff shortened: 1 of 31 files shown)


.github/workflows/pull-lokalise.yml
```diff
@@ -0,0 +1,91 @@
+name: Pull locales from Lokalise
+
+on:
+  workflow_dispatch:
+
+concurrency:
```
AndreaCimini90, line 6: The concurrency group depends on the selected ref, but every run writes to the same lokalise/translations branch. Two manual runs dispatched from different refs can therefore execute concurrently and overwrite or reject each other's commits. Use a repository-wide group such as lokalise-pull, or include the actual destination branch in the group.

.github/workflows/push-lokalise.yml
```diff
@@ -0,0 +1,86 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - develop
+    paths:
+      - 'packages/*/public/locales/it/**'
+
+concurrency:
```
AndreaCimini90, line 11: On push events, github.head_ref is empty, so the fallback to github.run_id gives every run a unique concurrency group. Consecutive pushes to develop can therefore upload concurrently, and an older run finishing last may overwrite newer translations—especially with rambo_mode and cleanup_mode. Use a stable group such as lokalise-push-${{ github.ref }} and consider canceling stale runs.

.github/workflows/pull-lokalise.yml
```diff
@@ -0,0 +1,91 @@
...
+          "PG": "pn-personagiuridica-webapp"
+        }
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          # L'action deve poter risolvere il branch base e riusare il branch di lavoro
+          # se esiste gia' sul remoto: serve la storia completa, non lo shallow clone.
+          fetch-depth: 0
+      - name: Pull ${{ matrix.app }} locales from Lokalise
```
AndreaCimini90, line 53: workflow_dispatch can be started from any branch, while the action derives both the PR base and Lokalise tag filter from the triggering ref. Running it from a feature branch can therefore create/update the shared translation branch against the wrong base and download a different key set. Explicitly set override_base_branch: develop and fail early when github.ref_name != 'develop', unless non-develop runs are intentionally supported.

.github/workflows/push-lokalise.yml
```diff
@@ -0,0 +1,86 @@
...
+  # write serve al primo run per creare/pushare il tag 'lokalise-upload-complete'.
+  # Dopo che il tag esiste si puo' tornare a 'contents: read' (come io-app).
+  contents: write
+
+jobs:
+  # Ogni app ha il suo progetto Lokalise, quindi qui rileviamo quali app hanno
+  # traduzioni italiane modificate. L'output 'changes' di paths-filter e' gia' un
+  # array JSON con i nomi dei filtri che hanno fatto match: lo passiamo tale e quale
+  # come matrix del job successivo, cosi' le app non toccate non generano un job.
+  detect-changes:
```
AndreaCimini90: This PR introduces stateful workflows that can update four external projects and delete remote keys through cleanup_mode, but no automated validation or execution evidence is included. Please add workflow tests/static checks for dispatch and push matrices, or document successful dry runs proving app detection, sequential pull commits, PR reuse, and cleanup behavior.

AndreaCimini90 (changes requested):

.github/workflows/pull-lokalise.yml
fbianchicodermine: Good catch. The single `lokalise/translations` branch is intentional: Lokalise is the single source of truth, so one pull run must produce one PR covering all four apps. But you are right that the lock was more granular than the resource it protects, which defeated `max-parallel: 1`. The concurrency group is now repository-wide `lokalise-pull`, so dispatches from different refs queue instead of running together.

.github/workflows/push-lokalise.yml
fbianchicodermine: Confirmed, `github.head_ref` is only populated on `pull_request` events, so on `push` and `workflow_dispatch` the group fell back to `run_id` and was unique per run. The group is now `lokalise-push`, repository-wide rather than per ref, since the resource to protect is the Lokalise project and that is the same whatever the ref is. `cancel-in-progress` stays `false`.

Related change in the same file: the detection job is gone, and the push now always runs the full `[PFL, PF, PA, PG]` matrix.

.github/workflows/pull-lokalise.yml
fbianchicodermine: Addressed, splitting it in two.

Tag filter: the pull now sets `skip_include_tags: true` and downloads every key of the project. Each app has its own Lokalise project, so there is nothing left for the tag to isolate, and the previous behaviour was a trap. Keys are tagged with the branch that pushed them, and the pull action terminates when no key carries the tag of the triggering branch, so after merging this PR the first pull from `develop` would have downloaded nothing, silently.

PR base: added `override_base_branch: develop`, plus a `check-branch` job that fails the run when `github.ref_name != 'develop'`, so the pull can only be dispatched from `develop`.

.github/workflows/push-lokalise.yml
fbianchicodermine: These workflows have no unit-testable surface, so both were validated by running them against the real Lokalise projects from this branch.

Pull, run [33860489802](https://github.com/pagopa/pn-frontend/actions/runs/33860489802): four jobs executed sequentially (PFL, PF, PA, PG), each committing onto the reused `lokalise/translations` branch, producing the single PR #2028 with 96 files and four distinct commits, one per app. PR reuse across runs is visible on #2012 and #2023, both updated by later runs instead of spawning new PRs.

Push, run [33859809520](https://github.com/pagopa/pn-frontend/actions/runs/33859809520): app detection correctly isolated PA, PF and PG and skipped PFL, and all three uploads ran with `cleanup_mode: true` (that detection step is now removed, see the concurrency thread).

AndreaCimini90 (approved):

